Issue #42, opened 2018-07-18
Change the `GameController` to display the appropriate views


tomclose:
Currently the `GameController` displays a simple waiting page for most steps:
```js
<WaitingStep message="confirmation of adjudicator deployment" />
```
This doesn't make sense in terms of the game. At the very least it should probably do something like:

- [ ] After the play has been chosen, that choice should display on each of the screens
- [ ] When the opponent's play is chosen, that should be displayed along with the outcome of the game
- [ ] At the end of the game there should be an option to play again and make another choice
- [ ] On top of this, we should work towards making it feel a bit more like the experience you'd want a user to have when playing the game.

PR #50, opened 2018-07-21
Add in ready to send states

### LINKED ISSUE #45: Add in the `ReadyToSend*` states
The [Game Controller](https://github.com/magmo/rps-poc/blob/master/src/components/GameController.jsx) doesn't currently contain any of the `ReadyToSend*` states, skipping straight to the `WaitingFor*` instead. We should add these in. Suggested behaviour:

* User takes an action e.g. chooses their play
* App transitions to the `ReadyToSendPropose` state, and displays e.g. the previous screen with "...sending" at the bottom
* The `messageSender` detects the transition and auto-sends the message and emits the `messageSent` action
* The `gameReducer` detects `messageSent`, calls `gameEngine.messageSent()`, which returns `WaitingForAccept`, which becomes the new application state

To make this workflow clear to the user, we might need to insert a delay in the `messageSender` to ensure that this state always lasts e.g. 1 sec. I think we can do this with sagas by combining `all` with `delay`.

tomclose:
Fixes #45

### DIFF
--- a/src/components/GameController.jsx
+++ b/src/components/GameController.jsx
@@ -15,8 +15,14 @@ export default class GameController extends PureComponent {
     } = this.props;
 
     switch (applicationState && applicationState.type) {
+      case playerAStates.ReadyToSendPreFundSetup0:
+        return <WaitingStep message="ready to propose game" />;
+
       case playerAStates.WaitForPreFundSetup1:
         return <WaitingStep message="opponent to accept game" />;
+
+      case playerAStates.ReadyToDeploy:
+        return <WaitingStep message="ready to deploy adjudicator" />;
 
       case playerAStates.WaitForBlockchainDeploy:
         return <WaitingStep message="confirmation of adjudicator deployment" />;
@@ -24,14 +30,23 @@ export default class GameController extends PureComponent {
       case playerAStates.WaitForBToDeposit:
         return <WaitingStep message="confirmation of opponent's deposit" />;
 
+      case playerAStates.ReadyToSendPostFundSetup0:
+        return <WaitingStep message="ready to send deposit confirmation" />;
+
       case playerAStates.WaitForPostFundSetup1:
         return <WaitingStep message="opponent to confirm deposits" />;
 
       case playerAStates.ReadyToChooseAPlay:
         return <SelectPlayStep handleSelectPlay={chooseAPlay} />;
 
+      case playerAStates.ReadyToSendPropose:
+        return <WaitingStep message="ready to send round proposal" />;
+
       case playerAStates.WaitForAccept:
         return <WaitingStep message="opponent to choose their move" />;
+
+      case playerAStates.ReadyToSendReveal:
+        return <WaitingStep message="ready to send reveal" />;
 
       case playerAStates.WaitForResting:
         return <WaitingStep message="opponent to accept the outcome" />;
--- a/src/components/HomePage.jsx
+++ b/src/components/HomePage.jsx
@@ -37,5 +37,3 @@ export default function HomePage({ loginUser }) {
     </div>
   );
 }
-
-export default HomePage;
--- a/src/game-engine/application-states/ApplicationStatesPlayerA.js
+++ b/src/game-engine/application-states/ApplicationStatesPlayerA.js
@@ -36,6 +36,10 @@ class BasePlayerA {
             stake: this.stake,
         }
     }
+
+    get alwaysSend() {
+        return false;
+    }
 }
 
 class ReadyToSendPreFundSetup0 extends BasePlayerA {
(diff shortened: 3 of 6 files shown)


src/components/GameController.jsx
```diff
@@ -15,24 +15,39 @@ export default class GameController extends PureComponent {
     } = this.props;
 
     switch (applicationState && applicationState.type) {
+      case playerAStates.ReadyToSendPreFundSetup0:
+        return <WaitingStep message="ready to propose game" />;
```
amhunt: Eventually I still think we'd want to have a mapping from `applicationState` to `viewComponent` so that we don't need so many `<WaitingStep />` repetitions. It's fine for now, but especially if we get rid of the messages in the future (which I'm assuming we will?) then we should probably move away from this giant `switch`

src/components/HomePage.jsx
```diff
@@ -36,6 +36,4 @@ export default function HomePage({ loginUser }) {
       </div>
     </div>
   );
-}
-
-export default HomePage;
\ No newline at end of file
+}
```
amhunt: I've seen this on a lot of your PRs - it'd be nice to configure your editor to leave a newline at the end (this is pretty standard for JS).

src/redux/sagas/index.js
```diff
@@ -1,12 +1,14 @@
...
+import { call, put, takeEvery, select, all } from 'redux-saga/effects';
 import { authProvider, reduxSagaFirebase } from '../../gateways/firebase';
-import { types, loginSuccess, messageReceived, eventReceived } from '../actions';
+import { types, loginSuccess, messageReceived, eventReceived, messageSent } from '../actions';
 
 function* messageSender() {
   let state = yield select();
-  if (state.message) {
+  if (state.alwaysSend) {
+    yield delay(2000);  // for dev purposes
```
amhunt: why are all of these yield statements? I'm not very familiar with this pattern, but it seems like this would require calling `next()` four times on the generator function to get to the end of this function? I'm probably just not understanding, but maybe you could add a comment to explain?

src/redux/reducers/game.js
```diff
@@ -24,11 +24,31 @@ export default function gameReducer(state = {}, action = {}) {
...
+      return new playerA.ReadyToSendPropose({ ...coreProps, adjudicator, aPlay, salt, signedProposeMessage });
+
+    case types.MESSAGE_SENT:
+      switch (state.type) {
+        case playerA.types.ReadyToSendPreFundSetup0:
+          let signedPreFundSetup0Message = "blah";
+          return new playerA.WaitForPreFundSetup1({ ...coreProps, signedPreFundSetup0Message });
+        case playerA.types.ReadyToSendPostFundSetup0:
+          let signedPostFundSetup0Message = "blah";
+          return new playerA.WaitForPostFundSetup1({ channel, stake, balances, adjudicator, signedPostFundSetup0Message });
```
amhunt: `...coreProps` here as well?

src/redux/reducers/game.js
```diff
@@ -52,7 +72,7 @@ export default function gameReducer(state = {}, action = {}) {
           return new playerA.WaitForBToDeposit({ channel, stake, balances, adjudicator });
         case playerA.types.WaitForBToDeposit:
           let signedPostFundSetup0Message = "blah";
-          return new playerA.WaitForPostFundSetup1({ channel, stake, balances, adjudicator, signedPostFundSetup0Message });
+          return new playerA.ReadyToSendPostFundSetup0({ channel, stake, balances, adjudicator, signedPostFundSetup0Message });
```
amhunt: `coreProps` here

src/game-engine/application-states/ApplicationStatesPlayerA.js
```diff
@@ -36,6 +36,10 @@ class BasePlayerA {
             stake: this.stake,
         }
     }
+
+    get alwaysSend() {
```
amhunt: why is this called `alwaysSend`?  Why not just `shouldSendMessage`? I guess I'm wondering what this implies... that if `alwaysSend()` returns false, then we should still send a message sometimes?

src/components/GameController.jsx
tomclose: I think I agree with this. We'll definitely want to swap out the `<WatingStep />`s for the appropriate views (see https://github.com/magmo/rps-poc/issues/42) first. After that we can probably do a good refactor.

src/components/HomePage.jsx
tomclose: Done. In VSCode you add `"files.insertFinalNewline": true,` to your user settings cc @andrewgordstewart .

src/redux/sagas/index.js
tomclose: This is the redux-saga [effects](https://redux-saga.js.org/docs/basics/Effect.html) pattern. Every effect is yielded to the code that runs the saga, which then manages how these effects are performed. I think this makes things like cancelling actions (and all their dependent actions) easy.

src/game-engine/application-states/ApplicationStatesPlayerA.js
tomclose: Changed.

PR #51, opened 2018-07-23
Firebase opponents

tomclose:
Uses redux-firebase-saga to load the list of opponents and keep it synced.

### DIFF
--- a/src/redux/sagas/opponents.js
+++ b/src/redux/sagas/opponents.js
@@ -1,0 +1,33 @@
+import { fork, take, cancel } from 'redux-saga/effects';
+import { reduxSagaFirebase } from '../../gateways/firebase';
+
+import {
+  types,
+  syncOpponents,
+} from '../actions/opponents';
+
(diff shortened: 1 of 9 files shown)


src/redux/sagas/opponents.js
```diff
@@ -0,0 +1,33 @@
...
+    'opponents',
+    {
+      successActionCreator: syncOpponents,
+      transform: opponentsTransformer,
+    }
+  );
+}
+
+export default function * opponentSaga () {
+  while( yield take(types.OPPONENTS.SUBSCRIBE) ) {
```
andrewgordstewart: Not sure I fully understand, but doesn't the `yield` keyword always return a truthy value?

>The yield keyword actually returns an IteratorResult object with two properties ...

tomclose: Yep. The think is it doesn’t necessarily return that value immediately - the loop here only runs when a new event is received.

PR #58, opened 2018-07-29
Moved message send to index saga

n-white:
Added logic to the saga to send message to firebase. 

Also, updated the `OpponentsSelectionStep` with the `messageSent` action. 

Future PR todos:
- Still need to add more code later to actually generate the preFundMessage.
- Also, need to add saga in later PR for listening for opponent messages.

### DIFF
--- a/src/redux/sagas/index.js
+++ b/src/redux/sagas/index.js
@@ -1,26 +1,33 @@
 import { delay } from 'redux-saga';
+import {
+  put, takeEvery, select, fork, call,
+} from 'redux-saga/effects';
 import { types as playerAStates } from '../../game-engine/application-states/ApplicationStatesPlayerA';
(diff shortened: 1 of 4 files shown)


src/redux/sagas/index.js
```diff
@@ -1,27 +1,34 @@
...
+}
 
 function* messageSender() {
   let state = yield select(getApplicationState);
   if (state.shouldSendMessage) {
     yield delay(2000);  // for dev purposes
     yield put(messageSent(state.message));
-    // todo put the message sending logic here
+    // push to firebase messages - organized by channel ID
+    yield call(reduxSagaFirebase.create, `messages.${state.channel}`, {
```
tomclose: Should this be `state.channelId`?

PR #63, opened 2018-08-01
Strip out RPSGame and RPSState

tomclose:
This PR replaces `RPSGame` and `RPSState` with the new 'pledge' approach. It also converts more of the game engine files to typescript.


src/game-engine/GameEngine.ts
```diff
@@ -154,36 +147,34 @@ export default class GameEngine {
...
 
       // The opponent's state assumes that B won
-      const { resolution } = opponentState;
-      if (result === RpsGame.Results.TIE) {
-        resolution[0] += opponentState.stake;
-        resolution[1] -= opponentState.stake;
-      } else if (result === RpsGame.Results.A) {
-        resolution[0] += 2 * opponentState.stake;
-        resolution[1] -= 2 * opponentState.stake;
+      let newBalances = balances;
```
amhunt: I think this will only create a shallow copy - probably want to use `[...balances]` for safety.

PR #67, opened 2018-08-03
sync messages from db

n-white:
Setup saga / action / reducer for syncing messages from firebase.

Todos:
- sync messages for only the current channel
- subscribe to messages during game play
- unsubscribe from messages outside of game play

### DIFF
--- a/src/redux/actions/messages.js
+++ b/src/redux/actions/messages.js
@@ -1,0 +1,15 @@
+import makeEnum from '../lib/make_enum';
+
+export const types = makeEnum([
+  'SYNC_MESSAGES',
+  'SUBSCRIBE_MESSAGES',
+]);
+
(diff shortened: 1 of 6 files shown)


src/redux/actions/messages.js
```diff
@@ -0,0 +1,15 @@
+import makeEnum from '../lib/make_enum';
+
+export const types = makeEnum([
+  'SYNC_MESSAGES',
```
n-white: @tomclose happy to follow the nested convention you were using elsewhere, if you prefer that (like `types.MESSAGES.SYNC`. Seemed unnecessary to have it nested for messages so did it this way.

tomclose:
Going to merge this as is - we can do the message filtering in a later PR.

PR #74, opened 2018-08-09
Application controller as a saga

tomclose:


src/game-engine/application-states/PlayerA/ReadyToSendPreFundSetupA.ts
```diff
@@ -1,25 +1,26 @@
...
   toJSON() {
     return {
       ...this.commonAttributes,
-      message: this.message,
+      move: this.move,
     };
   }
 
-  get shouldSendMessage() {
+  get shouldSendMove() {
```
andrewgordstewart: ✂️  ?

src/game-engine/application-states/PlayerA/ReadyToSendPropose.ts
```diff
@@ -1,22 +1,23 @@
...
     super({ channel, stake, balances });
     this.aPlay = aPlay;
     this.salt = salt;
-    this.message = message;
+    this.move = move;
     this.adjudicator = adjudicator;
   }
 
-  get shouldSendMessage() {
+  get shouldSendMove() {
```
andrewgordstewart: ✂️ ?

src/redux/sagas/messages.ts
```diff
@@ -1,24 +1,44 @@
...
-    {
-      successActionCreator: MessageAction.syncMessages,
-    },
-    'value'
+
+// watch messages on a given address
+
+// ? send messages to a given address
+
+function * sendMessage(action: any) {
```
andrewgordstewart: I suppose we could make a `sendMessageAction` interface just above, specifying the `to` and `data` members?

tomclose: I'm confused about how best to do this. We _should_ already have the type from the action, but I don't seem to be able to access it (as in `MessageAction.sendMessage` isn't actually a type). I think this is because we bundle all the types together into `MessageAction` but can't access them separately. We probably need to alter how we do the types of actions to fix this ...

tomclose: As in, for the proper fix.

PR #75, opened 2018-08-09
[WIP] Jsx files are now Tsx files

andrewgordstewart:


src/components/GameController.tsx
```diff
@@ -6,7 +6,23 @@ import WaitingStep from './WaitingStep';
 import SelectPlayStep from './SelectPlayStep';
 import * as playerAStates from '../game-engine/application-states/PlayerA';
 
-export default class GameController extends PureComponent {
+interface Props {
+  applicationState: playerAStates.PlayerAState;
+  chooseAPlay: any; // TODO: what should this be?
```
andrewgordstewart: Not sure what `chooseAPlay` is meant to do.

src/components/WaitForOpponentStep.tsx
```diff
@@ -18,12 +22,10 @@ export default class WaitForOpponentStep extends React.PureComponent {
         </div>
         {selectedPlayId && (
           <div style={{ width: '100%' }}>
-            You&apos;ve chosen {PLAY_OPTIONS.find(option => option.id === selectedPlayId).name}
+            You&apos;ve chosen {Play[selectedPlayId]}
```
andrewgordstewart: I don't know how/if we should ensure that `selectedPlayId` is valid.

src/components/WaitForOpponentStep.tsx
```diff
@@ -18,12 +22,10 @@ export default class WaitForOpponentStep extends React.PureComponent {
         </div>
         {selectedPlayId && (
```
andrewgordstewart: We can replace `selectedPlayId &&` with `Play[selectedPlayId] &&` on line 18 to deal with the below comment

src/components/GameController.tsx
tomclose: That should be the function that triggers the action.

src/components/GameController.tsx
```diff
@@ -60,15 +76,20 @@ export default class GameController extends PureComponent {
           handleMessageSent={messageSent}
           handleCreateChallenge={chooseOpponent}
           opponents={opponents}
+          handleSelectChallenge={chooseAPlay} // TODO: right?
```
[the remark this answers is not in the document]
andrewgordstewart: I guess it shouldn't be...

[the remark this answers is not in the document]
tomclose: No, I don't think so. We need to rework a lot of this stuff though, so this will be ok for now.

src/components/WaitForOpponentStep.tsx
tomclose: I don't think we actually need to convert to and id and back. We should be able to iterate over the `Plays`, with the button for a given play triggering the action with that play baked in.

src/components/GameController.tsx
```diff
@@ -23,6 +23,14 @@ interface Props {
 }
 
 export default class GameController extends PureComponent<Props> {
+  static propTypes = {
+    applicationState: PropTypes.object.isRequired,
+    chooseAPlay: PropTypes.func.isRequired,
+    chooseOpponent: PropTypes.func.isRequired,
+    opponents: PropTypes.arrayOf(PropTypes.object).isRequired,
+    subscribeOpponents: PropTypes.func.isRequired,
+  };
```
tomclose: Do we still need `propTypes` if we have the `Props` interface defined?

PR #76, opened 2018-08-10
Opponent selection working again

tomclose:


src/components/GameController.tsx
```diff
@@ -1,42 +1,26 @@
...
 import WaitingStep from './WaitingStep';
 import SelectPlayStep from './SelectPlayStep';
 import * as playerAStates from '../game-engine/application-states/PlayerA';
+import { GameState } from '../redux/reducers/game';
+import { Opponent } from '../redux/reducers/opponents';
 
 interface Props {
-  applicationState: playerAStates.PlayerAState;
+  applicationState: GameState;
   chooseAPlay: any; // TODO: what should this be?
```
andrewgordstewart: Can we fix this? Is `chooseAPlay` even used anymore?

src/components/GameController.tsx
```diff
@@ -81,10 +65,8 @@ export default class GameController extends PureComponent<Props> {
       default:
         subscribeOpponents();
         return <OpponentSelectionStep
-          handleMessageSent={messageSent}
-          handleCreateChallenge={chooseOpponent}
+          chooseOpponent={chooseOpponent}
```
andrewgordstewart: 👍

src/components/GameController.tsx
```diff
@@ -1,42 +1,26 @@
...
 import SelectPlayStep from './SelectPlayStep';
 import * as playerAStates from '../game-engine/application-states/PlayerA';
+import { GameState } from '../redux/reducers/game';
+import { Opponent } from '../redux/reducers/opponents';
 
 interface Props {
-  applicationState: playerAStates.PlayerAState;
+  applicationState: GameState;
   chooseAPlay: any; // TODO: what should this be?
   chooseOpponent: (x: any, y: any) => void;
```
andrewgordstewart: Are we able to specify the types of `x` and `y`, and choose better names?

src/redux/actions/opponents.ts
```diff
@@ -7,7 +7,7 @@ export enum OpponentActionType {
 };
 
 export const OpponentAction = {
-  syncOpponents: (opponents: string[]) => ({
+  syncOpponents: (opponents: any[]) => ({
```
andrewgordstewart: Why not
`import { Opponent } from '$ROOT_DIR/src/reducers/opponents';`
and then 
`syncOpponents: (opponents: Opponent[]) => ({`
?

src/redux/reducers/opponents.ts
```diff
@@ -1,15 +1,20 @@
 import { Reducer } from 'redux';
 import { OpponentActionType, OpponentAction } from '../actions/opponents';
 
-export interface OpponentState {
-  opponents: string[];
+export interface Opponent {
```
andrewgordstewart: Is this the right file to define an `Opponent` interface?j

src/components/GameController.tsx
andrewgordstewart: Ditto for the other functions

README.md
```diff
@@ -1,13 +1,5 @@
 ## Rock Paper Scissors: A Force-Move Games POC
 
-**NOTE:** the app currently requires to be run against the [generate-types](https://github.com/magmo/force-move-games/tree/generate-types)
```
andrewgordstewart: Side note -- since we'll likely continue to update `fmg-core`, and might not publish new versions straight away, it might be a worthy investment to make a more robust development process. For example, perhaps we'll be working on a new branch in `rps-poc` that requires some changes in `fmg-core`, but we want to develop the changes in parallel.

One can
1. Check that the latest published version of `fmg-core` is up to date with master/some-user-inputted (?) branch.
2a. If so, continue.
2b. If not,
  - if running on circle-ci, throw an error. tests fail.
  - if running locally, a script links the correct local version for you. tests function as intended.

src/redux/reducers/opponents.ts
tomclose: Probably not. I think it'll do for now though - anticipating a re-org later when we have a better overview of the project.

README.md
tomclose: One of the disadvantages of a mono-repo is you can't do an npm git link for a package stored inside a mono-repo. Otherwise that approach would solve this.

PR #79, opened 2018-08-15
Add player B states

tomclose:
This PR adds the `playerBstates` to the game controller, so that screens will display for them. There's still a lot of work to be done here but this PR puts the basic structure in place.

It also renames the `ChooseAPlay` action to `ChoosePlay`, so that it can be used for both players.

### DIFF
--- a/src/components/GameController.tsx
+++ b/src/components/GameController.tsx
@@ -83,33 +83,33 @@ export default class GameController extends PureComponent<Props> {
         return <WaitingStep message="Wait for resting" />;
       
       case playerB.ReadyToSendPreFundSetupB:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="ready to send prefund setup" />;
 
       case playerB.WaitForAToDeploy:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="waiting for adjudicator to be deployed" />;
 
       case playerB.ReadyToDeposit:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="ready to deposit funds" />;
 
       case playerB.WaitForBlockchainDeposit:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="waiting for deposit confirmation" />;
 
       case playerB.WaitForPostFundSetupA:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="waiting for post fund setup" />;
 
       case playerB.ReadyToSendPostFundSetupB:
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="ready to send post fund setup" />;
 
       case playerB.ReadyToChooseBPlay:
         return <SelectPlayStep choosePlay={choosePlay} />;
 
       case playerB.ReadyToSendAccept:
         // your choice
-        return <WaitingStep message="opponent to accept the outcome" />;
+        return <WaitingStep message="ready to send accept" />;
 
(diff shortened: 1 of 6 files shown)


src/components/SelectPlayStep.tsx
```diff
@@ -24,7 +24,7 @@ export default class SelectPlayStep extends React.PureComponent<Props> {
           {Object.keys(Play).filter(key => !isNaN(Number(key))).map(option => (
             <button
               type="button"
-              onClick={() => chooseAPlay(parseInt(option,10))}
+              onClick={() => choosePlay(Play[option])}
```
tomclose: This seems to work (?)

lalexgap: Hmm when I had `Play[option]` I was seeing this in the debugger:
image

src/components/GameController.tsx
```diff
@@ -41,40 +42,77 @@ export default class GameController extends PureComponent<Props> {
     }
 
     switch (applicationState && applicationState.constructor) {
-      case playerAStates.ReadyToSendPreFundSetupA:
+      case playerA.ReadyToSendPreFundSetupA:
```
lalexgap: It'd be nice if we could pull out some of this switch logic so the component isn't so verbose. Maybe a simple mapper that maps a `state` -> `message`. Probably something for a later date when the UI evolves.

src/components/SelectPlayStep.tsx
tomclose: I think I'm pretty confused about enums in typescript. I'm surprised that the code compiles in typescript but then the it doesn't work:
```js
enum Number {
    Zero,
    One,
}

function difference(a: Number, b: Number) {
    return a - b;
}

console.log(difference(Number.One, Number.Zero)); //=> 1
console.log(difference(Number[1], Number[0])); //=> NaN
```
Why does this compile? ^

src/components/SelectPlayStep.tsx
```diff
@@ -8,8 +8,36 @@ interface Props {
...
+      </div>
+    </button>
+    )
+  }
+
+
   render() {
-    const { choosePlay, afterOpponent } = this.props;
+    const { afterOpponent, choosePlay } = this.props;
+    const renderChooseButton = this.renderChooseButton;
```
lalexgap: Just FYI this is why we couldn't access `this.props` in `renderChooseButton`.  It looks like when we do  `const renderChooseButton = this.renderChooseButton;` When `renderChooseButton`  executs `this` is the global context. See https://www.typescriptlang.org/play/#src=class%20Dummy%20%7B%0D%0A%20%20%20%20doStuff()%20%7B%0D%0A%20%20%20%20%20%20%20%20console.log(this)%3B%0D%0A%20%20%20%20%20%20%20%20console.log(this.data)%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20%20doOtherStuff()%20%7B%0D%0A%20%20%20%20%20%20%20%20%20const%20doStuff%20%3D%20this.doStuff%3B%0D%0A%20%20%20%20%20%20%20%20%20this.doStuff()%3B%0D%0A%20%20%20%20%20%20%20%20%20doStuff()%3B%0D%0A%20%20%20%20%7D%0D%0A%0D%0A%20%20%20%20data%20%3D%205%3B%0D%0A%20%20%20%20%0D%0A%7D%0D%0A%0D%0Aconst%20dummy%20%3D%20new%20Dummy()%3B%0D%0Adummy.doOtherStuff()%3B

We can switch `   { renderChooseButton(choosePlay, Play.Rock, "Rock") }` to    `{this. renderChooseButton(Play.Rock, "Rock") }`. That way in `renderChooseButton` `this` is the React Component.

PR #81, opened 2018-08-16
Added Typing to Actions

lalexgap:
Added typing to Actions so we can use the proper action type instead of `action:any`. It's a little verbose but yay type safety!


src/redux/reducers/login.ts
```diff
@@ -17,7 +17,7 @@ const initialState: LoginState = {
   player: undefined,
 };
 
-export const loginReducer: Reducer<LoginState> = (state = initialState, action) => {
+export const loginReducer: Reducer<LoginState> = (state = initialState, action:LoginAction) => {
```
tomclose: I also have this change in an open PR :p. When I did it I found I had to make some changes to the format of `LoginState` to avoid compilation errors: https://github.com/magmo/rps-poc/pull/79/commits/b895fa7ad2bc4161afdb4b5bd02e93f7918d02d8

tomclose: Oh I see what happened - you changed `player: string` on the `loginSuccess` action. The player actually needs to be an `object`.

src/redux/actions/game.ts
```diff
@@ -1,16 +1,15 @@
...
 
+
 export enum GameActionType {
   CHOOSE_OPPONENT = 'GAME.CHOOSE_OPPONENT',
   CHOOSE_A_PLAY = 'GAME.CHOOSE_A_PLAY',
   EVENT_RECEIVED = 'GAME.EVENT_RECEIVED',
   MOVE_RECEIVED = 'GAME.MOVE_RECEIVED',
   MOVE_SENT = 'GAME.MOVE_SENT',
-  STATE_CHANGED = 'GAME.STATE_CHANGED',
+  STATE_CHANGED = 'GAME.STATE_CHANGED'
```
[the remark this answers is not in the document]
lalexgap: Currently, our ts-lint is not set up to enforce this (otherwise it would fail to build). I can add the rule `"trailing-comma": [true, {"multiline":{"objects": "always"}}]` to tslint.config to enforce this.

src/redux/reducers/login.ts
lalexgap: Oops, will adjust.

PR #82, opened 2018-08-17
Add Basic Auto Opponent

lalexgap:
Added a simple auto opponent that will always play rock.


src/components/SelectPlayStep.tsx
```diff
@@ -24,7 +24,7 @@ export default class SelectPlayStep extends React.PureComponent<Props> {
           {Object.keys(Play).filter(key => !isNaN(Number(key))).map(option => (
             <button
               type="button"
-              onClick={() => chooseAPlay(Play[option])}
+              onClick={() => chooseAPlay(parseInt(option,10))}
```
lalexgap: The enum was not parsing properly before, so the compare logic would not work correctly. Probably a better way to do this.

src/redux/sagas/auto-opponent.ts
```diff
@@ -0,0 +1,79 @@
...
+      // Start up the game engine for our autoplayer B
+      gameEngine = fromProposal({ move: Move.fromHex(action.data), wallet });
+      yield put(MessageAction.messageReceived(gameEngine.state.move.toHex()));
+    } else {
+      switch (action.type) {
+        case MessageActionType.SEND_MESSAGE:
+          yield handleMessage(gameEngine);
+          break;
+
+        case GameActionType.STATE_CHANGED:
```
tomclose: This is a little bit confusing to read, as you have to look back to know that `STATE_CHANGED => state == WaitToDeposit` due to the filter. I'm not sure how to make this clearer though. Maybe add a comment?

src/redux/sagas/application-controller.ts
```diff
@@ -52,10 +53,19 @@ export default function* applicationControllerSaga(wallet: ChannelWallet) {
...
     if (newState) {
       if (newState.isReadyToSend) {
+
         yield put(MessageAction.sendMessage(newState.opponentAddress, newState.move.toHex()))
         yield put(GameAction.moveSent(newState.move));
       }
-      yield put(GameAction.stateChanged(newState));
+      // Fake sending a transaction for now
+      if (newState instanceof ReadyToDeploy && gameEngine){
+          gameEngine!.transactionSent();
```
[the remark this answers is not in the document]
lalexgap: Just a quick cheat to avoid another null check. `gameEngine!.transactionSent()` lets the compiler know that gameEngine will not be null, so it won't complain about it. Since this is just faking a transaction for now I figured it would be ok.

src/redux/sagas/auto-opponent.ts
lalexgap: Good point! We could change the filter to just catch all `STATE_CHANGED` actions and then have an `if` statement in the case. Would make it clearer but we'd be catching a ton more actions. Might just add a comment for now.

PR #84, opened 2018-08-18
Delays in auto-opponent saga

tomclose:
To allow us to see all the screens when developing.

### DIFF
--- a/src/redux/sagas/auto-opponent.ts
+++ b/src/redux/sagas/auto-opponent.ts
@@ -5,6 +5,7 @@ import {
   actionChannel,
   select,
 } from 'redux-saga/effects';
+import { delay } from 'redux-saga';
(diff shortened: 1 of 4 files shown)


src/redux/sagas/auto-opponent.ts
```diff
@@ -31,49 +31,53 @@ function* startAutoOpponent() {
...
-
-  if (newState instanceof ReadyToChooseBPlay) {
-    // Good ol rock, nothings beats that!
-    newState = gameEngine.choosePlay(Play.Rock);
-  }
+function* continueWithFollowingActions(gameEngine: GameEngine) {
+  while (true) { // keep going until we don't have an action to take
+    const state = gameEngine.state;
+    // tslint:disable-next-line:no-console
+    console.log(state);
```
lalexgap: Probably want to remove this :)

PR #89, opened 2018-08-20
Fix issue 86

andrewgordstewart:


tomclose:
This solves the issue but is more of a short-term fix. In the long-term we probably need an extra application state to account for this situation.

andrewgordstewart:
Agreed.

PR #90, opened 2018-08-21
Exclude current user from list of opponents

n-white:
It's weird for a user to be able to "challenge" themselves. Thought it would make more sense to exclude the current user from the list of opponents

image

### DIFF
--- a/src/containers/GameContainer.ts
+++ b/src/containers/GameContainer.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import { connect } from 'react-redux';
 
 import { GameAction } from '../redux/actions/game';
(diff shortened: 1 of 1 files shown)


src/containers/GameContainer.ts
```diff
@@ -1,14 +1,22 @@
...
-  opponents: state.opponents,
-});
+const mapStateToProps = (state: ApplicationState) => {
+	return {
+	  applicationState: state.game,
+	  opponents: getOpponents(),
+	}
+
+	function getOpponents() {
+		const currentUserAddress = _.get(state, 'login.wallet.account.address');
```
lalexgap: Once we have things typed properly, we could change this to directly access state.login.wallet.account.address.  Too bad `wallet` is `any` :(

PR #91, opened 2018-08-21
Improve the way we handle insufficient balances.

andrewgordstewart:
Rather than throw an error, transition to an InsufficientFunds state,
notifying the player that they should conclude the game or add more
funds (currently impossible with the simple adjudicator).


src/game-engine/GameEngineA.ts
```diff
@@ -147,10 +147,16 @@ export default class GameEngineA {
...
-
     const { balances, turnNum, stake, channel, adjudicator } = this.state;
 
+    if (this.state.stake > this.state.balances[0]) {
+      return this.transitionTo(
+        new State.InsufficientFundsA({
+          channel, balances, adjudicator,
+        })
+      )
+    }
```
tomclose: You're running this check at the point that A makes their choice. This made me wonder whether this is a good user experience, or whether we could let the user know earlier. It seems like we know in the resting state whether or not another round is possible or not. Do you think it would be preferable to transition to `InsufficentFundsA` instead of `ReadyToChooseAPlay`?

Also, should we check that B also has sufficient funds here?

src/game-engine/GameEngineB.ts
```diff
@@ -138,10 +138,16 @@ export default class GameEngineB {
...
-
     const { channel, stake, balances, preCommit, adjudicator, turnNum } = this.state;
 
+    if (this.state.stake > this.state.balances[1]) {
+      return this.transitionTo(
+        new State.InsufficientFundsB({
+          channel, balances, adjudicator,
+        })
+      )
+    }
```
tomclose: Same question as above - could we let the user know earlier?

src/game-engine/GameEngineA.ts
andrewgordstewart: I'm not sure about the UX, and took the lazy strategy of waiting for the last moment. It probably makes more sense to transition to `InsufficentFundsA` over `ReadyToChooseAPlay`, though, and notify the user as early as possible.

>Also, should we check that B also has sufficient funds here?
I'm open to this, but if we do, we need to add another view, and another state: `PlayerA.InsufficientFundsB`, or something like that.

src/game-engine/GameEngineB.ts
andrewgordstewart: I think if we're going to let the user know earlier, we should commit to that, and check that both players have enough stake to cover a round. If not, A isn't allowed to choose a move.

PR #93, opened 2018-08-21
Create app folder

andrewgordstewart:
With the goal of partitioning truffle-related files in its own folder, I've created an `app` folder, and put everything needed for the application to run in it, and nothing more.

As a later PR, we should bubble up some of the contents of `app`, but I've done it this way so that every single change was a simple rename. Getting this merged ASAP will minimize merge conflicts, and hence developer pain.


andrewgordstewart:
Nevermind!

PR #96, opened 2018-08-22
Move wallet into a seperate folder and decouple sagas

lalexgap:
- Moved wallet related stuff into the src/wallet subfolder. 
- Changed the wallet saga so it's a standalone saga and other sagas can use actions to communicate with it.
- Removed dependency on `Move` from `ChannelWallet`
- Added a wallet interface so the app doesn't have to know the details of `ChannelWallet`


src/wallet/sagas/wallet.ts
```diff
@@ -0,0 +1,74 @@
...
+  uid: string;
+  privateKey: string;
+  address: string;
+}
+
+export function* walletSaga(): IterableIterator<any> {
+  while (true) {
+    const action = yield take(WalletActionType.WALLET_REQUESTED);
+    const { uid } = action;
+    let wallet = yield* fetchWallet(uid);
```
tomclose: What does this `*` do?

src/game-engine/GameEngineA.ts
```diff
@@ -12,18 +11,19 @@ import Accept from './positions/Accept';
...
   static setupGame({ opponent, stake, balances, wallet }: 
-    { opponent: string, stake: number, balances: number[], wallet: ChannelWallet }
+    { opponent: string, stake: number, balances: number[], wallet: Wallet }
   ) {
     const participants = [wallet.address, opponent];
     const channel = new Channel(fakeGameLibraryAddress, 456, participants);
 
     const nextPledge = new PreFundSetup(channel, 0, balances, 0, stake);
-    const move = wallet.sign(nextPledge.toHex());
+    const move =new Move(nextPledge.toHex(), wallet.sign(nextPledge.toHex()));
```
tomclose: I think we might want `Move` to ultimately be part of the wallet interface. My initial thought was that the wallet would be responsible for taking position-strings from the app and returning `Moves` to be sent to the opponent, and then also for taking `Moves` from the opponent, checking that they're appropriately signed and returning position-strings to the app.

This is fine for now though - we can do all of that as part of a later refactor.

src/game-engine/GameEngineB.ts
```diff
@@ -10,9 +9,10 @@ import Propose from './positions/Propose';
 import Accept from './positions/Accept';
 import Resting from './positions/Resting';
 import Conclude from './positions/Conclude';
+import { Wallet } from '../wallet';
```
tomclose: I'm in two minds about whether we should keep `ChannelWallet` instead of `Wallet` to avoid confusion with things like meta-mask. This is an easy change to make later though, so happy with this here.

lalexgap: I agree `Wallet` is pretty generic, would be nice to have a better name

PR #97, opened 2018-08-22
Updated views

tomclose:
Begin to separate out the screens for each step.


src/components/PlaySelectedPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { Play } from '../game-engine/positions';
+
+interface Props {
+  message: string;
+  yourPlay: Play;
+}
+
+export default class PlaySelectedPage extends React.PureComponent<Props> {
+  static propTypes = {
```
andrewgordstewart: Are these still needed with typescript?

src/components/WaitingStep.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
```
andrewgordstewart: Should be deleted too?

PR #99, opened 2018-08-23
Separate out Wallet State/Engine

lalexgap:
Initial changes to separate out wallet state and create its own engine.

There's some further clean up to do and lots of tests to write.


lalexgap:
There should be more tests but I'd like to get this in first.

src/game-engine/GameEngineA.ts
```diff
@@ -104,46 +104,24 @@ export default class GameEngineA {
     }
   }
 
-  transactionSent(): State.PlayerAState {
-    if (!(this.state instanceof State.ReadyToDeploy)) { return this.state };
-
+  receiveFunding(event) {
```
tomclose: Maybe `fundingConfirmed`?

src/redux/sagas/application-controller.ts
```diff
@@ -43,28 +42,32 @@ export default function* applicationControllerSaga(wallet: Wallet) {
...
+      if (gameEngine != null) {
+        if (newState.isReadyToSend) {
+          yield put(MessageAction.sendMessage(newState.opponentAddress, newState.move.toHex()));
+          yield put(GameAction.moveSent(newState.move));
+        }
+        if (newState.isReadyForFunding) {
+          yield put(WalletFundingAction.walletFundingRequest(wallet, newState.player));
+          const fundingAction: WalletFundingAction = yield take(
+            WalletFundingActionType.WALLETFUNDING_FUNDED,
+          );
```
tomclose: It looks to me like this is going to block until the wallet returns (right?). I'm wondering whether this is what we want to do, as (a) the game state won't be updated and (b) the application controller will become blocked during this time. Would it work to handle the incoming `WALLETFUNDING_FUNDED` action above in the regular `appliction-controller` processing loop instead?

lalexgap: Good idea!

PR #103, opened 2018-08-24
Fix Issues with State flow

lalexgap:
Added an intermediate step so the application controller doesn't have to worry about whether funding has been sent. I'm seeing some errors from firebase so I'm going to hold off merging until those are resovled.


lalexgap:
Firebase errors were due to another PR

src/components/GameController.tsx
```diff
@@ -98,6 +101,10 @@ export default class GameController extends PureComponent<Props> {
 
       case playerB.ReadyToSendPreFundSetupB:
         return <WaitingStep message="ready to send prefund setup" />;
+      
+      case playerB.ReadyForFunding:
```
tomclose: `ReadyToFund`?

PR #105, opened 2018-08-24
Added authentication for for the play url

lalexgap:


src/components/PrivateRoute.tsx
```diff
@@ -0,0 +1,22 @@
+import { Route, Redirect } from 'react-router-dom';
+import React from 'react';
+
+const PrivateRoute = ({ component: Component, ...rest }) => (
+  <Route
+    {...rest}
+    render={props => {
+      return rest.isAuthenticated ? (
```
tomclose: I'm curious as to why you chose to bundle `isAuthenticated` up into `rest`, instead of having `isAuthenticated: Boolean` in the component arguments?

lalexgap: Mostly due to the sample I found doing it this way. But I think it would be better to explicitly have the isAuthenticated boolean in the component arguments. Will update.

PR #107, opened 2018-08-25
[WIP] Work on screens

tomclose:


src/wallet/components/WaitingStep.tsx
```diff
@@ -15,9 +15,11 @@ export default class WaitingStep extends React.PureComponent<Props> {
       <div className={css(styles.container)}>
         <div>
           <h1>
-            Waiting for {message}
-            ...
+            Wallet Screen
```
amhunt: I'm confused about this change. If this component is _called_ WaitingStep, it should probably involved something about Waiting, rather than just rendering the text `Wallet Screen`. If it's no longer specifically about waiting, then the name of the component should be changed.

src/components/GameController.tsx
```diff
@@ -55,10 +54,10 @@ export default class GameController extends PureComponent<Props> {
 
     switch (applicationState && applicationState.constructor) {
       case playerA.ReadyToSendPreFundSetupA:
-        return <ProposeGamePage message="ready to propose game" />;
+        return <GameProposedPage message="Waiting for opponent to accept game" />;
 
       case playerA.WaitForPreFundSetupB:
-        return <ProposalSentPage message="Waiting for opponent to accept game" />;
+        return <GameProposedPage message="Waiting for opponent to accept game" />;
```
amhunt: This no longer uses a message prop, so you should delete this prop (both in `GameProposedPage` and here)

tomclose: I think we probably will probably use this `message` in the longer term. The idea is for what's currently the `message` to be displayed in the status bar at the bottom of the page.

amhunt: Gotcha - I think the status bar will likely be something that is rendered by the GameController, not child components, but until we figure that structure out, I can leave it as a comment

amhunt:
@tomclose, still a bunch that needs to be done in terms of adding the functionality you show in your design mock, but here's what I did so far: 
- Implemented a basic fixed footer bar for the message (which should be refactored in the future to be a component rendered by `GameController`).
- Added a type for `currentPlayer` (and caught a potential bug in the process wow I love TS!)
- Fixed spacing on home page for buttons and centered icons

Feel free to merge and I can open up another PR later this week, or I can pick back up here.

### Footer bar
screen shot 2018-08-27 at 11 40 44 pm

### New button spacing
screen shot 2018-08-27 at 11 43 40 pm

PR #108, opened 2018-08-27
Configure project to work with truffle

andrewgordstewart:
Truffle seems to have some specific requirements on the flavour of javascript in the `truffle.js` config file, as well as the migrations used for contract deployment.

This PR keeps the truffle files as `ts`, maintaining the "no-javascript" rule (which is now enforced in `ts-config.json`). However, the syntax is changed to accommodate truffle, which breaks our linting rules. Thus, they're set to be ignored in `tslint.json`.

There are npm scripts to make it easy to compile, test and migrate our smart contracts.

WARNING: the truffle tests are totally broken. Because of that, `yarn truffle:compile` won't work right now. You can change it to `"truffle:compile": "tsc && tsc -p ./tsconfig.truffle.json ; run-s copy_truffle_files && (cd build/dist && ../../node_modules/.bin/truffle compile)",` to allow the script to continue past the broken 2nd step.

TODO:
- [ ] Extract the truffle scripts into shell files for readability.


package.json
```diff
@@ -58,27 +57,36 @@
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "node scripts/test.js --env=jsdom",
-    "postinstall": "rm -f node_modules/web3/index.d.ts"
+    "postinstall": "rm -f node_modules/web3/index.d.ts",
+    "copy_truffle_files": "copyfiles './contracts/**/*' './migrations/**/*' ./build/dist",
+    "truffle:compile": "(run-s copy_truffle_files && cd build/dist && ../../node_modules/.bin/truffle compile)",
+    "truffle:migrate": "(cd build/dist && ../../node_modules/.bin/truffle migrate)",
```
andrewgordstewart: This npm script doesn't let you specify the network.

.env.example
```diff
@@ -0,0 +1,7 @@
+ETH_ACCOUNT_MNENOMIC="a dozen random words"
+INFURA_API_KEY=<API_KEY>
```
tomclose: Where does this come from? Could you add a paragraph or two to the dev handbook?

package.json
```diff
@@ -58,27 +57,36 @@
...
+    "truffle:test": "(cd build/dist && ../../scripts/run-with-ganache.sh"
```
andrewgordstewart: The compiled tests currently don't run. Maybe the also have to be written in bad-style typescript?

package.json
```diff
@@ -58,27 +57,36 @@
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "node scripts/test.js --env=jsdom",
-    "postinstall": "rm -f node_modules/web3/index.d.ts"
+    "postinstall": "rm -f node_modules/web3/index.d.ts",
+    "copy_truffle_files": "copyfiles './contracts/**/*' './migrations/**/*' ./build/dist",
```
tomclose: How come this is `build/dist`, instead of e.g. `lib`?

andrewgordstewart: https://github.com/magmo/rps-poc/blame/master/tsconfig.json#L4

Either you or @amhunt set it this way. I think it was Andrew?

andrewgordstewart: Either way, it would be nice for this to not be a magic string, but I don't think you can access eg. environment variables inside `package.json`, right?

tomclose: I think this is fine here - `package.json` basically is a configuration file.

package.json
andrewgordstewart: Ok, they now run, thanks to separate configuration for truffle-related files. They fail miserably, however.

PR #109, opened 2018-08-29
Added drizzle configuration and deployment of contracts for Simple Adjudicator 

lalexgap:


config/webpack.config.dev.js
```diff
@@ -102,12 +102,6 @@ module.exports = {
       'react-native': 'react-native-web',
     },
     plugins: [
-      // Prevents users from importing files from outside of src/ (or node_modules/).
```
lalexgap: I removed this so we can import from outside the `src` folder to get the JSON contracts. It seemed cleaner then copying them into the `src` folder

migrations/2_deploy_contracts.ts
```diff
@@ -1,6 +1,8 @@
 var RockPaperScissorsGame = artifacts.require("./RockPaperScissorsGame.sol");
```
lalexgap: I'm not sure if we want to manually change the migration file but since `fmg` is separated into packages all their migration files use the same number/name.

src/redux/store.ts
```diff
@@ -1,22 +1,35 @@
+import { initDrizzle } from '../blockchain/drizzle';
 import { applyMiddleware, compose, createStore } from 'redux';
 import createSagaMiddleware from 'redux-saga';
+import { generateContractsInitialState } from 'drizzle';
 
+import RockPaperScissorsGame from '@contracts/RockPaperScissorsGame.json';
```
lalexgap: These references will fail until truffle:migrate is run. Not sure what the best way around that is.

src/blockchain/drizzle.ts
```diff
@@ -0,0 +1,9 @@
+import { Drizzle } from 'drizzle';
+let drizzle: Drizzle;
+export function initDrizzle(store: any, drizzleOptions: any):Drizzle {
+  drizzle = new Drizzle(drizzleOptions, store);
```
andrewgordstewart: `initDrizzle` should be idempotent, no?
if (drizzle) { return druzzle; }

src/redux/reducers/index.ts
```diff
@@ -4,17 +4,20 @@ import { gameReducer, GameState } from './game';
...
 import { loginReducer, LoginState } from './login';
 import { WalletState, walletStateReducer } from '../../wallet/redux/reducers/wallet-state';
+import { drizzleReducers } from 'drizzle';
 
 export interface ApplicationState {
   game: GameState;
   opponents: OpponentState;
   login: LoginState;
   wallet: WalletState;
+  contracts:any;
```
andrewgordstewart: `any[]` is slightly better, I suppose? Is there any way to define a `Contract` type?

src/redux/store.ts
andrewgordstewart: I think that's ok, it just needs to be part of the build process, right?

The annoying thing is that the contracts can't easily be watched and recompiled, but we won't often want to change the contracts (I hope)

tsconfig.json
```diff
@@ -14,7 +14,10 @@
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
     "noUnusedLocals": true,
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "paths": {
+      "@contracts/*": ["build/dist/build/contracts/*"]
```
andrewgordstewart: Do you want to just change the destination folder of `truffle:compile` instead?

src/redux/store.ts
andrewgordstewart: We also need to change the build process to strip out unnecessary information from the contracts, for when we actually want to deploy the app.

andrewgordstewart: 👍

migrations/2_deploy_contracts.ts
```diff
@@ -10,4 +12,11 @@ module.exports = function (deployer) {
   deployer.link(RockPaperScissorsState, RockPaperScissorsGame);
   deployer.link(State, RockPaperScissorsGame);
   deployer.deploy(RockPaperScissorsGame);
+
+  deployer.link(State, Rules);
+  deployer.deploy(Rules);
+  // Linking these without deploying doesn't seem to update the bytecode
+  // so we probably don't have to bother doing it.
+  deployer.link(State, SimpleAdjudicator);
+  deployer.link(Rules, SimpleAdjudicator);
```
tomclose: Linking these beforehand _does_ have an effect in the tests. There the contracts are loaded as:
```js
var RpsStateContract = artifacts.require("./RockPaperScissorsState.sol");
```
It could be that you're not getting the linked version because you're loading the file directly, instead of loading from the truffle "artifact area" - whatever that is.

PR #110, opened 2018-08-29
Fix smart contract tests

andrewgordstewart:
We didn't run truffle tests for a long time, completely ignoring the
smart contracts. The application architecture changed significantly, so
the tests are updated accordingly.

The npm test scripts are re-jiggered. Truffle tests are now run by circleCI. To run application tests in watch mode, run `yarn test:app` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "copy_truffle_files": "copyfiles './contracts/**/*' './migrations/**/*' ./build/dist",
     "copy_dotenv": "copyfiles '.env' ./build/dist",
     "truffle:compile": "tsc && tsc -p ./tsconfig.truffle.json && run-s copy_truffle_files && (cd build/dist && ../../node_modules/.bin/truffle compile)",
-    "truffle:migrate": "run-s truffle:compile copy_dotenv && (cd build/dist && ../../node_modules/.bin/truffle migrate --network $TRUFFLE_NETWORK)"
+    "truffle:migrate": "run-s truffle:compile copy_dotenv && (cd build/dist && ../../node_modules/.bin/truffle migrate --network ${TRUFFLE_NETWORK:development})"
   },
   "devDependencies": {
     "@babel/runtime": "^7.0.0-beta.53",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -56,13 +56,15 @@
...
+    "test:ci": "run-s test:truffle 'test:app --all --ci'",
+    "test:app": "node scripts/test.js --env=jsdom",
+    "test:truffle": "run-s truffle:compile && (cd build/dist && pwd && ../../scripts/run-with-ganache.sh)",
     "postinstall": "rm -f node_modules/web3/index.d.ts",
     "copy_truffle_files": "copyfiles './contracts/**/*' './migrations/**/*' ./build/dist",
     "copy_dotenv": "copyfiles '.env' ./build/dist",
     "truffle:compile": "tsc && tsc -p ./tsconfig.truffle.json && run-s copy_truffle_files && (cd build/dist && ../../node_modules/.bin/truffle compile)",
-    "truffle:migrate": "run-s truffle:compile copy_dotenv && (cd build/dist && ../../node_modules/.bin/truffle migrate --network $TRUFFLE_NETWORK)",
-    "truffle:test": "run-s truffle:compile && (cd build/dist && pwd && ../../scripts/run-with-ganache.sh)"
+    "truffle:migrate": "run-s truffle:compile copy_dotenv && (cd build/dist && ../../node_modules/.bin/truffle migrate --network $TRUFFLE_NETWORK)"
```
lalexgap: It might be nice to default this to the development network if `$TRUFFLE_NETWORK` is not set?

lalexgap:
We should probably update our tests to pull the contracts directly from the packages instead of depend on local files.

package.json
andrewgordstewart: Oh yeah, nice catch. I only didn't do this because I was lazy at the time :P

andrewgordstewart:
@lalexgap, `RockPaperScissorsGame` imports from the `fmg-core` package on npm, and not local files.

You might be confused, because there's a `fmg-rock-paper-scissors` package over there as well. But that package is on its way out, and should be removed.

PR #116, opened 2018-08-31
Refactor application states

tomclose:
Looking for early feedback on this:

* All states now have a `Position`
* The `Position` is a type parameter to the state, allowing us to alias properties in certain states where they exist.
* Also introduces a `PlayerAPositionType` enum, to replace using the `constructor.name` approach


src/game-engine/application-states/PlayerA/index.ts
```diff
@@ -1,50 +1,95 @@
...
+  readonly type = PlayerAStateType.WAIT_FOR_ACCEPT;
+  readonly isReadyToSend = false;
+
+  get aPlay() { return this.position.aPlay; }
+  get bPlay() { return this.position.bPlay; }
+  get salt() { return this.position.salt; }
+  get result() { return calculateResult(this.aPlay, this.bPlay); }
+}
+
+// todo: what should Position be here?
```
andrewgordstewart: Since the only option is to end the game, `CONCLUDE`?

src/components/GameController.tsx
```diff
@@ -55,100 +55,54 @@ export default class GameController extends PureComponent<Props> {
...
-      case playerA.ReadyToSendReveal:
-        const state0 = applicationState as playerA.ReadyToSendReveal;
-        return (
-          <ResultPage
-            message="Waiting for resting"
-            yourPlay={state0.aPlay}
-            theirPlay={state0.bPlay}
-            result={state0.result}
-          />
+          <PlaySelectedPage message="Waiting for opponent to accept" yourPlay={applicationState.aPlay} />
```
amhunt: Just a thought for later: these mapping make very little sense unless you are intimately familiar with the inner workings of the game engine. We will probably want to add comments or consider renaming some of these views.

src/components/GameController.tsx
```diff
@@ -55,100 +55,54 @@ export default class GameController extends PureComponent<Props> {
...
-      case playerA.InsufficientFundsB:
-        return (
-          <WaitingStep message="About to conclude the game – your opponent has run out of funds!" />
-        );
-
-      case playerB.ReadyToSendPreFundSetupB:
-        return <WaitingStep message="Ready to send prefund setup" />;
+      case playerA.INSUFFICIENT_FUNDS:
+        // todo: add into the logic about who it was that ran out of funds
+        return <WaitingStep message="About to conclude the game – either you or your opponent has run out of funds!" />;
```
amhunt: Why does `Insufficient Funds` correspond to `WaitingStep`? Is there actually something we are waiting for?

src/game-engine/GameEngineA.ts
```diff
@@ -178,33 +133,24 @@ export default class GameEngineA {
...
-    const { channel, stake, balances } = this.state;
+    const { stake, balances } = this.state;
 
-    if (this.state.stake > this.state.balances[0] || this.state.stake > this.state.balances[0]) {
-      return this.transitionTo(
-        new State.InsufficientFundsA({
-          channel, balances,
-        })
-      )
+    if (stake > balances[0] || stake > balances[0]) {
```
amhunt: this should be `if (stake > balances[0] || stake > balances[1]) {`
(the second `[0]` is currently incorrect)

src/game-engine/application-states/Base.ts
```diff
@@ -1,20 +1,34 @@
...
-  balances: number[];
-  channel: Channel;
-  stake: number;
-  type: string;
-  playerIndex: number; // overwritten by subclass
+import { Player } from '.';
+import { Position } from '../positions';
+
+export default class BaseState<T extends Position> {
+  readonly position: T;
```
amhunt: What is the `T` for? can't all of these just be `Position`?

src/game-engine/positions/PreFundSetupB.ts
```diff
@@ -0,0 +1,20 @@
+import { State } from 'fmg-core';
+import { packRestingAttributes } from '.';
+
+export default class PreFundSetup extends State {
+  stake: number;
+  turnNum: number;
+  stateCount: number;
+  balances: number[];
+  resolution: any;
```
amhunt: `number[]`?

src/redux/sagas/auto-opponent.ts
```diff
@@ -27,44 +27,36 @@ function* startAutoOpponent() {
...
-      // yield put (WalletFundingAction.walletFunded('0xComputerPlayerFakeAddress'));
-      gameEngine.fundingRequested();
-      const action = yield take(WalletFundingActionType.WALLETFUNDING_FUNDED);
-      gameEngine.fundingConfirmed({ adjudicator: action.adjudicator });
-    } else {
-      return false;
+    let state = gameEngine.state;
+
+    switch(state.type) {
+        case StateType.CHOOSE_PLAY:
```
amhunt: nit: looks like there are excess spaces here

src/components/GameController.tsx
tomclose: No - we just don't have a component for that yet, so are using `WaitingStep` temporarily.

src/game-engine/GameEngineA.ts
tomclose: Nice catch.

src/game-engine/application-states/Base.ts
tomclose: I had to do this to allow `ApplicationState` subclasses to be specialised to certain position subclasses, e.g:
```
export class WaitForResting extends BasePlayerA<Reveal> {
  readonly type = PlayerAStateType.WAIT_FOR_RESTING;
  readonly isReadyToSend = false;

  get stake() { return this.position.stake; }
  get aPlay() { return this.position.aPlay; }
  get bPlay() { return this.position.bPlay; }
  get salt() { return this.position.salt; }
  get result() { return calculateResult(this.aPlay, this.bPlay); }
}
```
If you can't specify that the `position` in the `WaitForResting` state is actually a `Reveal` position, you can't call e.g. `this.position.aPlay`, because not all `Positions` have an `aPlay` property.

src/components/GameController.tsx
amhunt: Great - just making sure :)

src/game-engine/application-states/Base.ts
amhunt: Ahhh, great - makes sense. Thanks for the explanation.

PR #117, opened 2018-09-02
Refactor wallet

tomclose:
This PR refactors the wallet to prepare for saving states. In particular it:

* Adds `VALIDATION_{REQUEST/SUCCESS/FAILURE}` actions, which will be used to the validate (and store) the positions sent by the opponent
* Adds `SIGNATURE_{REQUEST/SUCCESS/FAILURE}` actions, which will be used to sign (and store) the our positions before sending them to the opponent
* Updates the `FUNDING_{REQUEST/SUCCESS/FAILURE}` actions to follow the same pattern
* In all of the above uses a new actions pattern where all action type constants are exported directly. This seemed cleaner to write and use.
* Changes the initialization of the wallet: the `walletSaga` now takes care of the wallet initialization and is itself forked from the `loginSaga`, once the user id is known.
* The wallet address is communicated back to the application via a `initializationSuccess` action

This current version removes the existing blockchain interactions, stubbing out the funding step completely. I figure it will be easy to add this back in, now we know how that bit works.


src/redux/sagas/application-controller.ts
```diff
@@ -66,19 +65,20 @@ export default function* applicationControllerSaga(wallet: Wallet) {
...
         default:
           yield put(MessageAction.sendMessage(newState.opponentAddress, newState.position.toHex()));
       }
       yield put(GameAction.stateChanged(newState));
     }
   }
 }
+
+
+
```
amhunt: nit: remove these

src/redux/sagas/login.ts
```diff
@@ -40,19 +41,26 @@ function* loginStatusWatcherSaga() {
...
+
+      if (failure) { throw new Error('Wallet initialization timed out'); }
+
+      const address = (success as walletActions.InitializationSuccess).address;
+
+      const player = yield fetchOrCreatePlayer(address, user.displayName);
+
+      yield put(MessageAction.subscribeMessages(address));
+
+      applicationThread = yield fork(applicationControllerSaga, address);
```
amhunt: adding some comments here would be helpful (seems like a lot going on at first glance)

PR #120, opened 2018-09-04
presentation components for creating a new challenge

n-white:
Added components for creating a new challenge which is separate from the opponents table.

### DIFF
--- a/src/components/ChallengePage.tsx
+++ b/src/components/ChallengePage.tsx
@@ -1,0 +1,85 @@
+import _ from 'lodash';
+import React from 'react';
+
+import { Opponent } from '../redux/reducers/opponents';
+
+import CreateChallenge from './CreateChallenge';
+import OpponentSelectionStep from './OpponentSelectionStep';
+import Button from './Button';
+
+interface Props {
+	chooseOpponent: (opponentAddress: string, stake: number) => void;
+	playComputer: (stake: number) => void;
+	opponents: Opponent[];
+	currentPlayer?: {
+		address: string;
+		name: string;
+	};
+	createChallenge: (challenge: any) => void;
+}
+
+const initialState = { showChallenges: true };
+type State = Readonly<typeof initialState>;
+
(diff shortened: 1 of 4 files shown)


src/components/ChallengePage.tsx
```diff
@@ -0,0 +1,85 @@
...
+		address: string;
+		name: string;
+	};
+	createChallenge: (challenge: any) => void;
+}
+
+const initialState = { showChallenges: true };
+type State = Readonly<typeof initialState>;
+
+export default class ChallegePage extends React.PureComponent<Props, State> {
```
tomclose: typo: `Challenge`

src/components/ChallengePage.tsx
```diff
@@ -0,0 +1,85 @@
...
+		return (
+			<OpponentSelectionStep
+				chooseOpponent={this.props.chooseOpponent}
+				playComputer={this.props.playComputer}
+				opponents={this.props.opponents}
+				currentPlayer={this.props.currentPlayer}
+			/>
+		);
+	}
+}
```
tomclose: Seems like this file is indented with 8 spaces, instead of 2 ..?!

src/components/CreateChallenge.tsx
```diff
@@ -0,0 +1,49 @@
+import _ from 'lodash';
+import React from 'react';
+
+import Button from './Button';
+
+interface Props {
+  createChallenge: (challenge: any) => void;
+  currentPlayer?: {
```
tomclose: I think we know by this stage in the app that the `currentPlayer` must exist.

src/components/ChallengePage.tsx
tomclose: Nope .. it's tabs!

src/components/CreateChallenge.tsx
tomclose: This is a bigger change, as the parent component doesn't know for sure that `currentPlayer` exists ...

PR #123, opened 2018-09-04
Add simple adjudicator deployment

lalexgap:
Added a blockchain saga that will handle interacting with the blockchain. Added the ability to deploy the simple adjudicator to the blockchain as well as some crude error handling.

### DIFF
--- a/src/wallet/redux/actions/blockchain.ts
+++ b/src/wallet/redux/actions/blockchain.ts
@@ -1,8 +1,9 @@
-export type BlockchainDeployAdjudicatorAction = ReturnType<typeof deployAdjudicator>;
-export type BlockchainAdjudicatorDeployedAction = ReturnType<typeof adjudicatorDeployed>;
-export type BlockchainAction =
-  | BlockchainDeployAdjudicatorAction
-  | BlockchainAdjudicatorDeployedAction;
+export type DeploymentRequest = ReturnType<typeof deploymentRequest>;
+export type DeploymentSuccess = ReturnType<typeof deploymentSuccess>;
+export type DeploymentWrongNetworkFailure = ReturnType<typeof deploymentWrongNetworkFailure>;
+export type DeploymentMetamaskFailure = ReturnType<typeof deploymentMetamaskFailure>;
+export type DeploymentFailure = DeploymentWrongNetworkFailure | DeploymentMetamaskFailure;
+export type DeploymentResponse = DeploymentSuccess | DeploymentFailure;
 
 export const BLOCKCHAIN_DEPLOYADJUDICATOR = 'BLOCKCHAIN.DEPLOYADJUDICATOR';
 export const BLOCKCHAIN_ADJUDICATORDEPLOYED = 'BLOCKCHAIN.ADJUDICATORDEPLOYED';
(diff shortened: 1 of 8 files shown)


src/wallet/redux/actions/blockchain.ts
```diff
@@ -1,22 +1,33 @@
-import { Wallet } from '../..';
+export type BlockchainDeployAdjudicatorAction = ReturnType<typeof deployAdjudicator>;
```
tomclose: Maybe call this `DeployAdjudicator` to stick with the new action pattern (you'll get the `blockchain` part from the namespace). Potentially even better would be to have something like:
```ts
type DeploymentRequest = ReturnType<typeof deploymentRequest>;
type DeploymentSuccess = ReturnType<typeof deploymentSuccess>;
type DeploymentWrongNetworkFailure = ReturnType<typeof deploymentWrongNetworkFailure>;
type DeploymentMetamaskFailure = ReturnType<typeof deploymentMetamaskFailure>;
type DeploymentFailure = DeploymentWrongNetworkFailure | DeploymentMetamaskFailure;
type DeploymentResponse = DeploymentSuccess | DeploymentFailure;
```

src/wallet/redux/sagas/blockchain.ts
```diff
@@ -0,0 +1,34 @@
...
+// @ts-ignore
+import simpleAdjudicatorArtifact from 'fmg-simple-adjudicator/contracts/SimpleAdjudicator.sol';
+import contract from 'truffle-contract';
+import detectNetwork from 'web3-detect-network';
+// The blockchain saga will be responsible for dealing with the blockchain using truffle
+export function* blockchainSaga() {
+  const channel = yield actionChannel(blockchainActions.BLOCKCHAIN_DEPLOYADJUDICATOR);
+  while (true) {
+    const action = yield take(channel);
+    switch (action.type) {
```
tomclose: I'm guessing the switch is here in preparation for adding more types?

src/wallet/redux/sagas/blockchain.ts
```diff
@@ -0,0 +1,34 @@
...
+      case blockchainActions.BLOCKCHAIN_DEPLOYADJUDICATOR:
+        const simpleAdjudicatorContract = contract(simpleAdjudicatorArtifact);
+
+        const network = yield detectNetwork(web3.currentProvider);
```
tomclose: Should this be `yield call(detectNetwork, web3.currentProvider)`?

src/wallet/redux/actions/blockchain.ts
tomclose: Not entirely sure about the `Deployment*Failure` - maybe those are more general error actions. It's still useful to have that `DeploymentFailure` union type though, as you can then check you've covered all possibilities by using [the exhaustive checking pattern](https://github.com/magmo/rps-poc/blob/master/src/wallet/redux/sagas/wallet.ts#L38).

src/wallet/redux/sagas/blockchain.ts
lalexgap: Yup

src/wallet/redux/sagas/blockchain.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (!Object.keys(simpleAdjudicatorContract.networks).find(id => id === network.id)) {
```
tomclose: This is interesting. I'm guessing this means that we're actually shipping an app running on multiple networks?

PR #127, opened 2018-09-05
Start script will now honour .env values if they are present

lalexgap:


scripts/start.js
```diff
@@ -17,6 +14,11 @@ process.on('unhandledRejection', err => {
 // Ensure environment variables are read.
 require('../config/env');
 
+process.env.DEV_GANACHE_HOST = (process.env.DEV_GANACHE_HOST === undefined) ? 'localhost' : process.env.DEV_GANACHE_HOST;
```
kerzhner: @lalexgap do you think we need the explicit `=== undefined` here? If `DEV_GANACHE_HOST` is null or an empty string, do we still want to set it to the default `localhost`? If so, we could just use `process.env.DEV_GANACHE_HOST = (process.env.DEV_GANACHE_HOST) ? 'localhost' : process.env.DEV_GANACHE_HOST;`

scripts/start.js
```diff
@@ -100,15 +102,22 @@ choosePort(HOST, DEFAULT_PORT)
...
+    console.log(`Starting ganache on port ${process.env.DEV_GANACHE_PORT}`);
+    var ganacheServer = ganache.server({ port: process.env.DEV_GANACHE_PORT, network_id: 0, accounts });
+    ganacheServer.on('error', function (err) {
+      if (err.code && err.code==='EADDRINUSE') {
+        console.log(`Port ${process.env.DEV_GANACHE_PORT} in use. Assuming a ganache instance on that port.`);
+      } else {
+        throw err;
+      }
+    });
+    ganacheServer.listen(process.env.DEV_GANACHE_PORT, process.env.DEV_GANACHE_HOST, function (err, blockchain) {
```
kerzhner: @lalexgap if there is already another server instance running on the port, what happens here? Is this line needed?

lalexgap: Nope, it should only be called if the port is not in use

scripts/start.js
lalexgap: javascript doesn't like `process.env.DEV_GANACHE_HOST = (process.env.DEV_GANACHE_HOST) ? 'localhost' : process.env.DEV_GANACHE_HOST;` but I've changed it to be cleaner and use the `||` syntax which should catch empty/null values as well

PR #128, opened 2018-09-05
Update test script to fail on error and remove obsolete .gitingore entry

lalexgap:


scripts/test_truffle.js
```diff
@@ -22,9 +22,17 @@ try {
     }
   });
 
-  exec('truffle.cmd test --network ganache', (err, stdout, stderr) => {
+  exec('truffle test --network ganache', (err, stdout, stderr) => {
```
andrewgordstewart: Should this be using the project's specified version of truffle? ie. `node_modules/.bin/truffle`?

scripts/test_truffle.js
```diff
@@ -22,9 +22,17 @@ try {
...
     // Errors seem to be piped to stdout so we just output that always
```
andrewgordstewart: Comment is no longer true.

lalexgap: Oops

scripts/test_truffle.js
lalexgap: Yup, that's a good point.

PR #130, opened 2018-09-05
Remove drizzle

lalexgap:
Since we're planning on subscribing to contract events, we don't need to make use of drizzle. `Truffle-Contract` provides all the functionality we need.


tomclose:
Could you explain more about how that works? Does our app still have to poll for the contract events in some way?

PR #131, opened 2018-09-06
Build against test network

lalexgap:
I've deployed the contracts to the ropsten test network and altered the build script to reference the artifacts created by that migration.

The idea is that the various deployed contract artifacts will live in the `contracts/artifacts` folder. When building you'll be able to specify which network you're targetting which will link in the appropriate artifacts for that network.


src/registerServiceWorker.ts
```diff
@@ -34,7 +34,9 @@ export default function register() {
     }
 
     window.addEventListener('load', () => {
-      const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`;
+      // workaround for this bug https://github.com/wmonk/create-react-app-typescript/issues/346
```
lalexgap: Looks like this is fixed in the latest version of uglifyjs2, will verify and update

scripts/build.js
```diff
@@ -14,6 +14,11 @@ process.on('unhandledRejection', err => {
 // Ensure environment variables are read.
 require('../config/env');
 
+// Default to ropsten test network
+process.env.TARGET_NETWORK = process.env.TARGET_NETWORK || 'ropsten';
+process.env.DEFAULT_GAS = process.env.DEFAULT_GAS || 6721975;
+process.env.DEFAULT_GAS_PRICE = process.env.DEFAULT_GAS_PRICE || 20000000000;
```
tomclose: Might be worth adding instructions to readme on how to get ropsten eth.

lalexgap: Good point, I've updated the readme a little bit.

andrewgordstewart: Note that this is also already covered in the developer's handbook: https://github.com/magmo/developer-handbook/blob/master/deploying-contracts-with-truffle.md

PR #133, opened 2018-09-07
Deploying and depositing funds on the blockchain

lalexgap:
Allow for interaction with the blockchain to deploy the adjudicator and deposit funds. Still a little rough but wanted to get some eyes on it early.


src/utils/messages.ts
```diff
@@ -0,0 +1,31 @@
...
+    this.body = body;
+  }
+
+  get isWallet(): boolean {
+    return this.queue === Queue.WALLET;
+  }
+}
+
+export function decodeMessage(message: string) : Message {
+  const [queueName, body] = message.split('-');
```
tomclose: It would probably be easier to add the `queueName` as a property in firebase. That said, I don't think that should be a blocker for this PR.

src/wallet/redux/sagas/blockchain.ts
```diff
@@ -1,43 +1,76 @@
...
+  return channel;
+}
+
 // The blockchain saga will be responsible for dealing with the blockchain using truffle
 export function* blockchainSaga() {
   const channel = yield actionChannel(
     a =>
-      a.type === blockchainActions.DEPLOY_REQUEST ||
-      a.type === blockchainActions.DEPOSIT_REQUEST,
+      a.type === blockchainActions.DEPLOY_REQUEST || a.type === blockchainActions.DEPOSIT_REQUEST,
```
tomclose: Pretty sure you can do
```ts
const channel = yield actionChannel([
  blockchainActions.DEPLOY_REQUEST,
  blockchainActions.DEPLOSIT_REQUEST,
]);
```

lalexgap: That's much nicer!

PR #137, opened 2018-09-07
Fix up opponent selection page

andrewgordstewart:
This makes the page slightly more sensible, and also fixes a bug where challenges got added to the "players" node in firebase.


tomclose:
This is also fixed in https://github.com/magmo/rps-poc/pull/138.

PR #138, opened 2018-09-07
Refactor redux

tomclose:
This PR refactors the redux portions of the app.

* The `redux` folder is now organised by area (e.g. `message-service`) instead of function (e.g. `saga`).
* The are three major 'rooms' (`lobby`, `waiting-room`, `game`) with cleaner transitions between them.
* The auto-opponent now intercepts messages at the `message-service` level, preventing those messages from being pushed to firebase.
* In all other ways, the auto-opponent appears and functions like a regular opponent.
* In bringing the wallet messaging across, I've added a `queue` property to the message itself, instead of encoding in the data.
* I've cleared up some unused views.
* I've stopped the auto-opponent from getting stuck at the end of the game.

I haven't been able to test the wallet portion of this, as the login didn't seem to work for me in browsers other than chrome. Also I ran out of steam ...


src/components/SelectChallenge.tsx
```diff
@@ -34,20 +30,24 @@ export default class OpponentSelectionStep extends React.PureComponent<Props> {
...
               </tr>
-              {opponents.map(opponent => (
-                <tr key={opponent.address}>
-                  <td>{opponent.name}</td>
-                  <td>{opponent.wager}</td>
+              <tr key={autoOpponentAddress}>
+                <td>RockBot 🤖 </td>
+                <td>30</td>
+                <td>
+                  <Button onClick={() => acceptChallenge(autoOpponentAddress, 30)}>Challenge</Button>
```
andrewgordstewart: It's possible to accept a challenge from yourself, which results in a `SEND_MESSAGE` action followed by a `MESSAGE_RECEIVED` action.

self-selection

andrewgordstewart: I also can't quite figure out how to accept a challenge.

In firefox, I'm signed in using andrew.gord.stewart@gmail.com
In chrome, I'm signed in using andrew@magmo.com

In firefox, I click on the challenge created by the magmo user.
In chrome, I receive a message, but the received message isn't handled.

propose-fail

src/components/Header.tsx
```diff
@@ -1,28 +1,14 @@
...
 
-import { BRAND_COLOR, ROUTE_PATHS } from '../constants';
+import { BRAND_COLOR } from '../constants';
 
 export default function Header() {
   return (
     <div className={css(styles.container)}>
-      <div className={css(styles.leftNav)}>
-        <Link to="/">
-          <h3>RPS</h3>
```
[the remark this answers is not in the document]
tomclose: Yeah - this moves to having the whole app under a single url, where the page displayed is completely determined by the application state.

src/components/SelectChallenge.tsx
tomclose: So the challenge stuff needs some work. The idea is that you can only accept challenges from the `waiting-room`, which prevents you accepting your own challenges - hence the hanging when you do. If you leave the `waiting-room` through the app, your challenge will be deleted but if the app crashes or you refresh the page, it remains. To solve these problems, as well as the scenario where the user shuts their laptop, we could use a "heartbeat" approach, where `lastSeen` is updated every 10secs while in the `waiting-room` and the select opponent screen only displays challenges updated in the last 10secs (or so).

tomclose: I think what you're seeing in the gif is a bug. I wasn't able to test this as I could only get one browser working at once :(.

tomclose: Actually, on closer inspection it isn't - to accept the challenge you need to be in the `waiting-room` page - e.g. which you can only get to by creating a new challenge. Try this:

1. andrew@magmo.com creates a new challenge and stays on the waiting page
2. andrew.gord.stewart@gmail.com accepts that challenge

andrewgordstewart: so, based on the mechanism we use to accept challenges, challenges should only exist while you're in the "Waiting for someone to accept your challenge" page

the issue is that, if you're on that page, and you then leave the app (eg. you refresh) then the challenge sits there in perpetuity.

The only way around it I've thought is something like:
- firebase stores a `expiresAt` timestamp on a challenge
- when you're in the `waiting-room`, you send a message to firebase every so often, updating the `expiresAt` field to keep the challenge active.

andrewgordstewart: Sorry @tomclose for completely ignoring [your comment](https://github.com/magmo/rps-poc/pull/138#discussion_r216020010). It won't happen again.

andrewgordstewart: >The idea is that you can only accept challenges from the waiting-room, which prevents you accepting your own challenges

Another edge case to sort out is whether you can be signed into the app from two different locations. This approach wouldn't take care of that edge case.

src/redux/auto-opponent/saga.ts
```diff
@@ -0,0 +1,52 @@
+import { put, take, actionChannel } from 'redux-saga/effects';
+import { delay } from 'redux-saga';
+
+import * as autoOpponentActions from './actions';
+
+import { fromProposal, GameEngine } from '../../game-engine/GameEngine';
```
andrewgordstewart: This is the only place that uses `fromProposal` now. Otherwise, `GameEngineB.fromProposal` is used directly. We should unify this.

src/containers/GameContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <ResultPage
+          message="Waiting for resting"
+          yourPlay={state.aPlay}
+          theirPlay={state.bPlay}
+          result={state.result}
+        />
+      );
+
+    case playerA.INSUFFICIENT_FUNDS:
+      // todo: add into the logic about who it was that ran out of funds
```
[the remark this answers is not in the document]
tomclose: I think I removed that in the PR when I was simplifying the game states. The thinking was that it's possible to deduce from the state's properties who has run out, so I was going condense two states into one state + convenience method. But I then didn't add the convenience method ... :(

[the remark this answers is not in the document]
andrewgordstewart: 👍

andrewgordstewart:
The bugs discussed in [this thread](https://github.com/magmo/rps-poc/pull/138#discussion_r216016275) obviously have to be fixed, but that's outside the scope of this PR.

andrewgordstewart:
I much prefer this way of organizing redux, except that it's slightly harder to tell the context based on the filename.

https://xkcd.com/1172/

PR #145, opened 2018-09-07
Add option to logout

tomclose:
Looks a bit ghetto but does the job.


tomclose:
Included in https://github.com/magmo/rps-poc/pull/146

Issue #143, opened 2018-09-07
Expired challenges persist


andrewgordstewart:
See discussion here: https://github.com/magmo/rps-poc/pull/138#discussion_r216017706

Because these dead challenges persist, the UI is a bit confusing. I can challenge someone who's not online.

andrewgordstewart:
Another, lazier strategy would be:
- The challenges exist in firebase until they're either cancelled by the initiator or cleaned up
- A service somewhere can perform a clean-up: it checks that the initiator is in the waiting room. if not, it's deleted

I think that service could be a [cloud function](https://firebase.google.com/docs/functions/use-cases#perform_sanitization_and_maintenance)

andrewgordstewart:
Fixed by https://github.com/magmo/rps-poc/pull/154

Issue #142, opened 2018-09-07
You can challenge yourself.


andrewgordstewart:
You shouldn't be able to challenge yourself.

![self-selection](https://user-images.githubusercontent.com/8432675/45231304-d42fd100-b288-11e8-8900-f34f88c54b73.gif)

PR #147, opened 2018-09-07
Challenges expire after 5 seconds.

andrewgordstewart:
This almost takes care of https://github.com/magmo/rps-poc/issues/143 and https://github.com/magmo/rps-poc/issues/142

TODO: add a debounced sync action to take care of the mentioned edge case.

### DIFF
--- a/src/redux/lobby/saga.ts
+++ b/src/redux/lobby/saga.ts
@@ -46,9 +46,24 @@ export default function* lobbySaga(address: string) {
 };
 
 // maps { '0xabc': challenge1Data, ... } to [challenge1Data, ....]
-const challengeTransformer = (dict) => Object.keys(dict.value).map((key) => dict.value[key]);
+const challengeTransformer = (dict) => {
+  return Object.keys(dict.value).map((key) => {
+    return dict.value[key];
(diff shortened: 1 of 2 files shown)


src/redux/lobby/saga.ts
```diff
@@ -46,9 +46,24 @@ export default function* lobbySaga(address: string) {
 };
 
 // maps { '0xabc': challenge1Data, ... } to [challenge1Data, ....]
-const challengeTransformer = (dict) => Object.keys(dict.value).map((key) => dict.value[key]);
+const challengeTransformer = (dict) => {
+  return Object.keys(dict.value).map((key) => {
+    return dict.value[key];
+  }).filter((challenge) => {
+    return challenge.expiresAt > Date.now().toFixed()
```
tomclose: It might be worth adding a similar condition to the firebase query, to restrict the data you sync to those challenges that were active at the time you entered the lobby. It won't help with the accuracy but it would reduce the impact as the number of orphaned challenges increase.

andrewgordstewart: 👍

PR #149, opened 2018-09-07
More small improvements

tomclose:


src/containers/GameContainer.tsx
```diff
@@ -72,13 +72,26 @@ function GameContainer(props: GameProps) {
     case playerB.WAIT_FOR_POST_FUND_SETUP:
       return <WaitingStep message="Waiting for post-fund setup" />;
 
+    case playerB.WAIT_FOR_PROPOSE:
+      return <WaitingStep message="Waiting for your opponent to choose their move" />;
```
andrewgordstewart: Maybe make this message a constant, to make it consistent with [L47](https://github.com/magmo/rps-poc/pull/149/commits/7abd9ef2eb6eddba564fa1ee030462d8978631c0#diff-b6bedf4266977e58b7f3dba791a3f4a5R47)?

tomclose: My current thinking is that we'll do a complete overhaul of all this stuff when we have more design input. I'm expecting the message-passed-through approach to disappear completely.

PR #150, opened 2018-09-08
Switch over to BN.js

lalexgap:


lalexgap:
This should be good to go, I just want to go over it one more time before merging.

package.json
```diff
@@ -18,13 +18,14 @@
...
     "babel-runtime": "^6.23.0",
+    "bn.js": "^4.11.8",
     "case-sensitive-paths-webpack-plugin": "2.1.1",
     "chalk": "1.1.3",
     "css-loader": "0.28.7",
     "dotenv-expand": "4.2.0",
     "enum": "^2.5.0",
     "firebase": "^5.0.4",
-    "fmg-core": "^0.0.8",
+    "fmg-core": "0.0",
```
tomclose: typo

src/game-engine/positions/decode.test.ts
```diff
@@ -33,7 +34,9 @@ const testEncodeDecode = (pledge) => {
   it(`${pledge.constructor.name} is the same after encoding and decoding`, () => {
     const encoded = pledge.toHex();
     const decoded = decode(encoded);
-    expect(decoded).toEqual(pledge);
+    // We need to use JSON stringify due to the BN.js having possible different 
+    // internal representations of the same number
+    expect(JSON.stringify(decoded)).toEqual(JSON.stringify(pledge));
```
tomclose: This seems a bit worrying to me. Does this mean we can't reliably do `a === b` in the code?

lalexgap: I think either way we want to avoid using `===` for objects as in Javascript that is checking if they're the same reference not have the same values (which is what we're interested in here). I think this issue will mostly come up in testing as the `expect.areEqual` does a pretty deep equality comparison that can fail on BN.js numbers.

It might aslo fail on https://lodash.com/docs/4.17.10#isEqual as well

PR #152, opened 2018-09-10
Added a check for metamask on page load

### LINKED ISSUE #151: We should check for the metamask and the correct network on start
We should check that metamask is installed and the user has selected the correct network when the application is started. If they haven't we can provide instructions on what to do.

SInce switching networks in metamask refreshes the page we should only have to worry about this at the application start.

lalexgap:
- Added a check for metamask on page load
- Added a simple loader to show while the check is being performed

closes #151


lalexgap:
connects #151

PR #154, opened 2018-09-10
Debounced challenge expiration

andrewgordstewart:
This PR triggers a debounced expiration of challenges on every `SYNC_CHALLENGE` action.

### DIFF
--- a/src/redux/lobby/saga.ts
+++ b/src/redux/lobby/saga.ts
@@ -1,4 +1,4 @@
-import { fork, put, take, actionChannel, call, takeLatest, cancel } from 'redux-saga/effects'; 
+import { fork, put, take, actionChannel, call, takeLatest, } from 'redux-saga/effects'; 
 
 import { reduxSagaFirebase } from '../../gateways/firebase';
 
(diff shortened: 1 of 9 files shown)


src/redux/lobby/saga.ts
```diff
@@ -53,26 +55,35 @@ const challengeTransformer = (dict) => {
...
   );
+
+  yield takeLatest(lobbyActions.SYNC_CHALLENGES, expireChallenges)
 }
+
+function* expireChallenges() {
+  // This needs to be debounced at least as long as `CHALLENGE_EXPIRATION_INTERVAL`,
+  // in case we've just received a challenge that was just refreshed (In fact, this
+  // is the typical scenario.)
+  yield call(delay, CHALLENGE_EXPIRATION_INTERVAL)
```
andrewgordstewart: It would be nice to delay here just until the next-expiring-challenge, but I'm not sure how to get access to them without querying the database. It would also be a more complicated flow.

Perhaps it's alright to query just once to start off, and then debounce until it's necessary to expire some of the state's challenges?

PR #158, opened 2018-09-11
Added some basic error handling for interaction with the blockchain

lalexgap:
Handle errors interacting with the blockchain and show them to the user with the option to try again.

### DIFF
--- a/src/wallet/components/FundingError.tsx
+++ b/src/wallet/components/FundingError.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import Button from '../../components/Button';
+
+interface Props {
+  message: string;
+  tryAgain: () => void;
+}
+
+export default class FundingError extends React.PureComponent<Props> {
+  render() {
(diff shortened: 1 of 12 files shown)


src/wallet/components/FundingError.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface Props {
+  message: string;
+  tryAgain: () => void;
+}
+
+export default class FundingError extends React.PureComponent<Props> {
+  render() {
+    return <div>
+    <div>{this.props.message}</div>
+    <Button onClick={this.props.tryAgain}>Try Again</Button>
```
tomclose: Not important for this PR but in the long term we should have a "Give up" option, that would send a `FUNDING_FAILED` event to the app and return you there.

lalexgap: Yup, I agree. I figured we'd want that but I decided not to tackle it now.

PR #168, opened 2018-09-12
Refactor Wallet States and add Confirm Funding

lalexgap:
I've refactored the wallet states a bit to make them simpler and also introduced a simple wallet confirmation screen.


src/wallet/wallet-engine/positions/AdjudicatorReceived.ts
```diff
@@ -0,0 +1,6 @@
+export default class AdjudicatorReceived{
```
tomclose: This isn't particularly well explained, but in the main app the term `Position` is used to refer to the blockchain states - a position corresponds exactly to a state that the smart contract will accept. (This terminology has changed since we wrote the paper and `fmg-core`, where the same concept is referred to as `State` - but the intention is to go back and change those to make it consistent, once we're sure the new terminology is going to stick.) With this in mind, it might be better to call these objects in the wallet something different, as they don't share the specific meaning that `Position` has elsewhere. Maybe these could be states?

lalexgap:
Talking to @tomclose the way I'm approaching this is confusing. I'm going to close this one and create another after I fix some things up.

PR #169, opened 2018-09-13
Auto Deploy

lalexgap:
Created a build file for the google cloud build autodeploy.


lalexgap:
@tomclose Most of the configuration will be done in Google Cloud Build. We can select to build whenever a branch is tagged or create a deployment branch that whenever it's pushed to it deploys.

PR #174, opened 2018-09-13
Make the screens look a bit nicer

tomclose:
This is a short term fix to the styling so that we can start demo-ing this without being too embarrassed. I'm afraid that I've moved back to bootstrap, as pure.css was too big a learning curve and I needed to get it done quickly. I'm happy to remove bootstrap again in the longer term, when we have designs to work from and more time to implement.

styling


src/components/CreateChallenge.tsx
```diff
@@ -20,27 +21,49 @@ export default class CreateChallenge extends React.PureComponent<Props> {
...
+    this.nameInput.current.focus();
+  }
+
+  createChallengeHandler(e) {
+    e.preventDefault();
     const wager = Number(this.wagerInput.current.value);
     const name = String(this.nameInput.current.value);
 
+    // tslint:disable-next-line:no-console
+    console.log('in create challenge', wager, name);
```
lalexgap: Should this be removed?

src/components/ResultPage.tsx
```diff
@@ -20,37 +22,56 @@ export default class ResultPage extends React.PureComponent<Props> {
 
     switch (result) {
       case Result.YouWin:
-        return <h3> You won! </h3>;
+        return "You won! 🎉";
 
       case Result.YouLose:
-        return <h3> You lost </h3>;
+        return "You lost 😭";
```
lalexgap: 😭really sells the sour taste of defeat

PR #187, opened 2018-09-14
Create a seperate dev firebase DB

lalexgap:


config/webpack.config.dev.js
```diff
@@ -291,7 +291,10 @@ module.exports = {
       tsconfig: paths.appTsConfig,
       tslint: paths.appTsLint,
     }),
-  ],
+    new webpack.EnvironmentPlugin({
+    FIREBASE_PROJECT: 'rock-paper-scissors-dev',
```
andrewgordstewart: I don't see this project in my firebase console. I'm assuming it's created already?

lalexgap: Yup

PR #191, opened 2018-09-17
Sign and validate states when they are sent as messages

### LINKED ISSUE #176: The wallet should sign states
Currently the states we're sending from one player to the other aren't signed. Presumably this should work something like:

* In the `gameSaga` we alter `sendState` to send a `SIGNATURE_REQUEST` to the wallet and wait for a `SIGNATURE_SUCCESS` before calling `sendMessage`
* In the wallet we handle these, for the time being by just signing anything handed to it, and then sending the success message back

lalexgap:
We now sign and validate states as they are sent as messages. Decode logic for state from `fmg-core` has moved to the wallet while the game decode handles decoding game-specific state.  

We still need to handle signing wallet state.

closes #176

### DIFF
--- a/src/redux/game/saga.ts
+++ b/src/redux/game/saga.ts
@@ -32,9 +32,7 @@ export default function* gameSaga(gameEngine: GameEngine) {
 
     switch (action.type) {
       case messageActions.MESSAGE_RECEIVED:
-        yield put(walletActions.decodeStateRequest(action.message));
-        const decodeAction: walletActions.DecodeStateSuccess = yield take(walletActions.DECODE_STATE_SUCCESS);
-        newState = gameEngine.receivePosition(positionFromHex(decodeAction.state,action.message));
+        newState = gameEngine.receivePosition(positionFromHex(action.message));
         break;
       case gameActions.CHOOSE_PLAY:
         newState = gameEngine.choosePlay(action.play);
--- a/src/redux/game/saga.ts
+++ b/src/redux/game/saga.ts
@@ -32,7 +32,9 @@ export default function* gameSaga(gameEngine: GameEngine) {
 
     switch (action.type) {
       case messageActions.MESSAGE_RECEIVED:
-        newState = gameEngine.receivePosition(positionFromHex(action.message));
+        yield put(walletActions.decodeStateRequest(action.message));
+        const decodeAction: walletActions.DecodeStateSuccess = yield take(walletActions.DECODE_STATE_SUCCESS);
+        newState = gameEngine.receivePosition(positionFromHex(decodeAction.state,action.message));
         break;
       case gameActions.CHOOSE_PLAY:
         newState = gameEngine.choosePlay(action.play);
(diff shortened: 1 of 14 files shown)


src/redux/message-service/saga.ts
```diff
@@ -61,10 +69,44 @@ function* receiveFromFirebaseSaga(address: string) {
...
+  if (action.type === walletActions.VALIDATION_SUCCESS) {
+    return true;
+  } else {
+    // TODO: Properly handle this.
+    throw new Error("Signature Validation error");
+  }
+}
+
+function* signMessage(data) {
+  const requestId = hash(`${data}${Date.now()}`);
```
tomclose: Is this different to the `hash(data + Date.now())` used on line 78?

src/redux/message-service/saga.ts
```diff
@@ -61,10 +69,44 @@ function* receiveFromFirebaseSaga(address: string) {
   }
 }
 
+function* validateMessage(data, signature) {
+  const appState: ApplicationState = yield select(getApplicationState);
+  let opponentIndex = 0;
+  if (appState.gameState) {
+    opponentIndex = 1 - appState.gameState.playerIndex;
+  }
+  const requestId = hash(data + Date.now());
```
tomclose: This is a neat idea 👍

src/redux/game/saga.ts
```diff
@@ -32,7 +32,9 @@ export default function* gameSaga(gameEngine: GameEngine) {
 
     switch (action.type) {
       case messageActions.MESSAGE_RECEIVED:
-        newState = gameEngine.receivePosition(positionFromHex(action.message));
+        yield put(walletActions.decodeStateRequest(action.message));
+        const decodeAction: walletActions.DecodeStateSuccess = yield take(walletActions.DECODE_STATE_SUCCESS);
+        newState = gameEngine.receivePosition(positionFromHex(decodeAction.state,action.message));
```
tomclose: I see you've split the decoding between the wallet and the app, so that the wallet decodes the part of the state that it knows about and the app decodes the application-specific state. The way you've done this introduces a run-time dependency between these two components, so that now the app is unable to decode a state without calling out to the wallet. I wondered if it would be possible to remove this run-time dependency by instead making a code dependency between these two components e.g. the app imports `decode` from the wallet, allowing it to build its own app-state-aware `decode` method on top. This would potentially fit with a long-term pattern where the shared state decoding logic is contained in `fmg-core` and imported by both the wallet and the app. (It might even be worth refactoring to this now.)

src/redux/message-service/saga.ts
lalexgap: Nope, they should both be done the same way for consistency

src/redux/game/saga.ts
lalexgap: That's a good point. I'll put change the `game` state decode to just call through to the wallet state and put in an issue to pull it out to `fmg-core` proper.

PR #199, opened 2018-10-05
Game conclusions

andrewgordstewart:
This pull request allows users to conclude the game and withdraw funds.
There is still additional work to do on the conclusion UX
## TODO
* [x] implement conclusion logic for player B
* [x] get the blockchain listener to work in order to listen for the `GameConcluded` event, emitted by the SimpleAdjudicator
* [x] Add destination selection to the UI


tomclose:
Note about the `gameConcluded` event, as discussed with Alex: you always know that the `gameConcluded` event is coming, as both players must have signed a conclusion proof (and be on the wallet screen waiting) in order for the game to be concluded. It might be that we don't have to explicitly handle it here at all.

In general, there might not be a `gameConcluded` event when games conclude - as (when we move away from the simple adjudicator) a lot of games will be concluded completely off-chain.

src/game-engine/GameEngineA.ts
```diff
@@ -41,13 +41,13 @@ export default class GameEngineA {
...
   state: State.PlayerAState;
- 
+
   constructor(state) {
     this.state = state;
   }
 
   receivePosition(positionReceived: Position) {
-    switch(positionReceived.constructor) {
+    switch (positionReceived.constructor) {
```
tomclose: General point: I know we have this in our code a lot but I'm not convinced it is safe anymore (due to changes in names when minifying). We should probably figure out definitively if it is safe and move away from it if not.

src/redux/message-service/saga.ts
```diff
@@ -83,6 +88,7 @@ function* validateMessage(data, signature) {
     action = yield take(actionFilter);
   }
   if (action.type === walletActions.VALIDATION_SUCCESS) {
+    yield put(walletActions.storeMessageRequest(data, signature, Direction.Received));
```
tomclose: I think this should probably be done in the wallet e.g. if `VALIDATION_SUCCESS` comes back, you know that the wallet will have also stored it. I'm pretty sure there's never a case where you receive a valid state and don't want the wallet to save it, so we should remove the burden of remembering to save from the application developer.

src/redux/message-service/saga.ts
```diff
@@ -100,6 +106,8 @@ function* signMessage(data) {
   while (signatureResponse.requestId !== requestId) {
     signatureResponse = yield take(actionFilter);
   }
+
+  yield put(walletActions.storeMessageRequest(data, signatureResponse.signature, Direction.Sent));
```
tomclose: Same comment as above: this should probably be done inside the wallet as part of the signature request.

lalexgap: I think the reason it was originally split out was due to the fact that we want the wallet to be able to sign more than states (Like the address value that is sent for `withdraw`

tomclose: Wouldn't signing the final state always happen within the context of the wallet though? E.g. I open my channel wallet and instruct it to send the right transaction to metamask to withdraw to a given address (which will involve signing it). I might have missed other use cases though. What do you think?

PR #200, opened 2018-10-08
Handle timeout errors

andrewgordstewart:
The current behaviour actuall puts a `WALLET.INITIALIZATION.SUCCESS` action after the `WALLET.INITIALIZATION.FAILURE` action, and leaves you with a dead page.

image

### DIFF
--- a/src/components/ErrorPage.tsx
+++ b/src/components/ErrorPage.tsx
@@ -5,7 +5,7 @@ interface Props {
   error: string;
 }
 
-const initialState = { error: "Application Error" };
+const initialState = {};
 type State = Readonly<typeof initialState>;
 
 export default class ErrorPage extends React.PureComponent<Props, State> {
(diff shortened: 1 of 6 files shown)


src/components/ErrorPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default class ErrorPage extends React.PureComponent<Props, State> {
+  readonly state: State = initialState;
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return <div> Error: {this.state.error} </div>;
```
andrewgordstewart: I didn't quite get around to displaying the error message that was set by the reducer -- how do you do this again?

tomclose: The `error` is stored in the redux store and you've made that available by binding the store to the component in the `ErrorContainer`. All you need to do now is read the error from the props i.e. use `this.props.error`. You can remove the error from the state (and actually remove references to the state completely).

PR #203, opened 2018-10-10
Removing challenge heartbeats

kerzhner:
With this PR, any challenge entry in Firebase is a valid challenge entry.

Status quo of challenge validity:
1. After defining challenge properties, you enter the Waiting Room and create a Challenge entry in Firebase.
1. You periodically update the timestamp on the Challenge.
1. If you leave the Waiting room cleanly, the Challenge entry is removed.

With this PR:
1. After defining challenge properties, you enter the Waiting Room and create a Challenge entry in Firebase.
1. If you leave the Waiting room cleanly, the Challenge is removed.
1. If you abandon the Waiting room by navigating away from the best site on the internet https://demo.magmo.com, refreshing the page, or loosing the internet connection, then Firebase removes the challenge for you.


src/redux/waiting-room/saga.ts
```diff
@@ -31,55 +29,35 @@ export default function* waitingRoomSaga(
...
 
+  const challengeKey = `/challenges/${address}`;
+
   yield put(applicationActions.waitingRoomSuccess({ ...commonChallengeProps, stake }));
   const serializedChallenge = {
     ...commonChallengeProps,
     stake: stake.toString(),
   };
+  
+  yield firebase.database().ref(challengeKey).onDisconnect().remove();
```
tomclose: I _think_ you might need a `call` here for this to work reliably. Not 100% sure though.

Just to confirm, I'm guessing that this sends an instruction to firebase to remove the challenge on disconnect. Is that correct?

kerzhner: @tomclose Correct, this is an instruction to Firebase to delete the entry at the reference on client disconnect (https://firebase.google.com/docs/reference/js/firebase.database.OnDisconnect#remove).

I don't believe a `call` is necessary, but it is a good idea to use `call` for the code to be testable.

As a note, one flaw with this approach is that, if a client disconnects and re-connects while in the Lobby, Firebase deletes the challenge but does not create a new challenge on re-connect. There is a listener that allows an elegant solution to this flaw (`firebase.database().ref(".info/connected")`), but I am not sure it is worth our time at the moment to worry about this edge case.

Issue #180, opened 2018-09-13
Add the ability to launch a challenge on the blockchain


tomclose:
In the current app, it is not possible to call the `forceMove` operation. We need to add this functionality. Possible workflow:

* On each page of the app where you're waiting for your opponent, there's a button that says `launch challenge`
* This sends a `CHALLENGE_REQUEST` to the channel wallet and transitions to the channel wallet screen
* The channel wallet asks you to confirm, then sends the force-move transaction to metamask.
* The channel wallet waits until the transaction succeeds.
* You then stay in the channel wallet, until the challenge is complete. (In general, it's dangerous to continue playing, as if you provide an updated position to your opponent before the challenge is over, you can be refuted - and potentially punished..)

Things to consider:

* (In the long-run) we need to make sure that the app is still listening for incoming states while the user is navigating the create challenge screens - it could be that the state arrives just in time, so the challenge would be pointless. We could probably skip this edge-case for a v0 though.

Issue #181, opened 2018-09-13
Detecting and reacting to blockchain challenges


tomclose:
The channel wallet needs to be monitoring for adjudicator events, for the current channel. If a challenge is launched the wallet needs to halt the game and take over until the channel is resolved. Possible workflow:

* At the point the channel is funded, the wallet kicks off a monitoring saga that watches for events on the adjudicator address.
* If a force-move arrives, it send a some sort of "halt" message to the app. The should then pass control to the wallet.
* In the wallet, the user is presented with a possible course of action:
  * `conclude` if you already have a conclusion proof  
  * `refute` if the challenge state is stale
  * `respond` otherwise
  * [we won't handle `respondWithAlternative` here, as alternatives moves can't be generated from our app, so we will ignore them for the poc (although someone could write an alternative client that generates them but ¯\_(ツ)_/¯)\]
* the wallet waits until the transaction has gone through
* and, if successful, returns you to the app

PR #206, opened 2018-10-12
Challenges

andrewgordstewart:
This PR implements an issue-challenge workflow as well as a minimal detect-and-respond-to-challenges workflow.

## What works
Right screen: Resting
Left screen: Hit "Challenge"
Left screen: Submit transaction
Right screen: When transaction is recorded, wallet is displayed, prompting user to select "RespondWithMove" response.
Right screen: Submit transaction
Both screens: resume to normal play

With this workflow, this PR partially resolves https://github.com/magmo/rps-poc/issues/180 and https://github.com/magmo/rps-poc/issues/181.

## What doesn't work
Right screen: Resting
Left screen: Hit "Challenge"
Right screen: Choose move
Left screen: Submit transaction

At this point, Left screen has received the new move from Right Screen, and so they might as well not submit the transaction.

Once the transaction was recorded, the Refute and RespondWithAlternativeMove options show up, but they don't work.

## TODO
* [ ] The wallet should refuse to sign new states when a challenge has been detected
* [ ] The wallet should listen for incoming states, 
* [ ] The wallet should support `Refute` and `RespondWithAlternativeMove`
* [ ] The wallet should correctly compute the response options -- I don't think `RespondWithAlternativeMove` should be possible?


src/redux/game/saga.ts
```diff
@@ -69,28 +65,23 @@ export default function* gameSaga(gameEngine: GameEngine) {
...
-        opponentAddress,
-        myBalance,
-        opponentBalance,
-        playerIndex,
- ));
+    case PlayerBStateType.WAIT_FOR_POST_FUND_SETUP:
+      // Send the state to player A and then proceed with funding
+      yield sendState(state);
+      yield requestFunding(state);
+      yield sendState(state);
```
tomclose: Should there be two `sendState` calls here?

src/utils/randomHex.ts
```diff
@@ -0,0 +1,4 @@
+export function randomHex(n) {
+  // todo: make random
```
tomclose: 😄

PR #215, opened 2018-10-18
Fix issue 214

### LINKED ISSUE #214: Deployed contract promise doesn't resolve.
Commit: 5648247ae8a64d3c7c1a3ffbb9d3f7884ec7e06e
Location: https://github.com/magmo/rps-poc/blob/ag-challenge/src/wallet/redux/sagas/blockchain.ts#L40

Our app is (upredictably) victim to [this bug](https://github.com/trufflesuite/truffle-contract/issues/70#issuecomment-355376332), the obvious symptom being:
- upon successfully deploying the contract, the promise returned by `simpleAdjudicatorContract.new` does not resolve.
- since `call` blocks until that promise resolves, player B's wallet never gets the adjudicator's address, and thus Player A's does not as well
- consequently, both players' games are locked in a "waiting to deploy state"

andrewgordstewart:
Closes #214

### DIFF
--- a/src/contracts/SimpleAdjudicator.ts
+++ b/src/contracts/SimpleAdjudicator.ts
@@ -49,11 +49,11 @@ export async function deploySimpleAdjudicator({ channelId, amount }: { channelId
 }
 
 export async function simpleAdjudicatorAt({ address, amount } : { address: string, amount: BN }) {
-  const truffleContract = await setupContract();
+  const truffleContract = await setupContract(connectWeb3());
   return await truffleContract.at(address, { value: amount.toString() });
 }
 
-async function setupContract(connectedWeb3?) {
+async function setupContract(connectedWeb3) {
   const simpleAdjudicatorContract = contract(simpleAdjudicatorArtifact);
   simpleAdjudicatorContract.setProvider(connectedWeb3.currentProvider);
 
--- a/src/contracts/SimpleAdjudicator.ts
+++ b/src/contracts/SimpleAdjudicator.ts
@@ -9,13 +9,31 @@ import { connectWeb3 } from '../wallet/web3';
 export async function deploySimpleAdjudicator({ channelId, amount }: { channelId: string, amount: BN }) {
   const connectedWeb3 = connectWeb3();
   const truffleContract = await setupContract(connectedWeb3);
+  /*
+  Truffle's contract deployment workflow is not robust enough to deal with
+  metamask's fragility, so we manually deploy a contract using web3, grab
+  its address, and then instantiate a truffle contract at this address.
+
+  See: https://github.com/magmo/rps-poc/issues/214
+  */
 
   let contractBytecode;
   contractBytecode = truffleContract.bytecode;
 
   Object.keys(truffleContract.links).forEach(linkName => {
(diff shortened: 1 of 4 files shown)


src/contracts/SimpleAdjudicator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import connectedWeb3 from '../wallet/web3';
+
+export async function deploySimpleAdjudicator(channelId: string, amount: BN) {
+  const truffleContract = await setupContract();
+
+  let contractBytecode;
+  contractBytecode = truffleContract.bytecode;
+
+  Object.keys(truffleContract.links).forEach(linkName => {
```
lalexgap:
```suggestion
 // Manually link contracts into the byteCode so we can use web3 to deploy the bytecode
// This avoids an issue with truffleContract.new not always resolving.
 Object.keys(truffleContract.links).forEach(linkName => {
```

src/contracts/SimpleAdjudicator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import connectedWeb3 from '../wallet/web3';
+
+export async function deploySimpleAdjudicator({ channelId, amount }: { channelId: string, amount: BN }) {
+  const truffleContract = await setupContract();
+
+  let contractBytecode;
+  contractBytecode = truffleContract.bytecode;
+
+  Object.keys(truffleContract.links).forEach(linkName => {
```
kerzhner: @andrewgordstewart Maybe leave a comment on why these manipulations are necessary.

src/contracts/SimpleAdjudicator.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  const web3Contract = new connectedWeb3.eth.Contract(simpleAdjudicatorArtifact.abi, connectedWeb3.eth.defaultAccount);
+
+  const tx = web3Contract.deploy({
+    data: contractBytecode,
+    arguments: [channelId],
+  });
+
+  const deployedContract = await tx.send({ from: connectedWeb3.eth.defaultAccount, value: amount.toString() });
```
kerzhner: @andrewgordstewart Elsewhere in the app, we use the `yield` + the saga framework to block on asynchronous calls. Do you see us using the `await` pattern in other parts of the app going forward?

src/contracts/SimpleAdjudicator.ts
```diff
@@ -0,0 +1,46 @@
...
```
kerzhner: @andrewgordstewart Elsewhere in the app, we use the `yield` + the saga framework to block on asynchronous calls. Do you see us using the `await` pattern in other parts of the app going forward?

andrewgordstewart: I think this would improve testability, along with @tomclose's work on moving more logic into reducers.

Since reducers are synchronous, you couldn't move this logic into a reducer, for instance.

However, by moving this logic out of a saga, we could more easily write unit tests for, eg., `deployContract`.

So, I think we should explore this idea: extracting some async logic out of sagas, and using the (more common) `await` pattern.

Issue #212, opened 2018-10-17
Ability to run demo.magmo.com against other testnets


andrewgordstewart:
Right now, demo.magmo.com only runs against Ropsten.

Ropsten had been down for 4 days prior to this morning, so it would be good to be able to switch testnets (or even use the mainnet!).

* Ideally, the user could select from a list of available networks.
* Minimally, we should know how to redeploy the app against another test network. In this case, we should instruct the user which network they should use, if they're on the incorrect network.

Issue #156, opened 2018-09-10
Cannot run a dev server against predeployed contracts on a test network


lalexgap:
Currently when the build is run we build against already deployed contracts on a test network. However, when running the dev server we only run against ganache. We should provide an option to run the dev server against the already deployed contracts on the test network.

PR #216, opened 2018-10-19
Allow selecting a network

lalexgap:
adresses #212  and #156 

We now support building against multiple test networks. By default the prod build is ropsten and development is a local ganache instance.

However you can set the environment variable `TARGET_NETWORK` to override the defaults and to build against a target network.

I've also added support to run the dev server against a test net

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -22,6 +22,7 @@ const publicPath = '/';
 const publicUrl = '';
 // Get environment variables to inject into our app.
 const env = getClientEnvironment(publicUrl);
+const deployContractsFromLoader = filename=>process.env.TARGET_NETWORK==='development';
 
 // This is the development configuration.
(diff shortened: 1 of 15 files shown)


config/webpack.config.dev.js
```diff
@@ -22,6 +22,7 @@ const publicPath = '/';
 const publicUrl = '';
 // Get environment variables to inject into our app.
 const env = getClientEnvironment(publicUrl);
+const deployContractsFromLoader = filename=>process.env.TARGET_NETWORK==='development';
```
kerzhner: @lalexgap do we need to pass the `filename` argument to the function?

config/webpack.config.dev.js
```diff
@@ -137,7 +139,7 @@ module.exports = {
               }
             ]
           },
-          {
+          {
```
kerzhner: @lalexgap nit, unnecessary space here.

PR #225, opened 2018-10-24
WIP: Navigation Bar

kerzhner:
Opening an early PR for the navigation bar to get feedback. Specifically:
1. Do you think there is a more elegant way to add the navigation bar component to pages? At the moment, every page that uses the navigation bar will have to add `NavigationBar` in the `render` method.
1. Any thoughts on how we should structure our sass styling modules?


kerzhner:
@lalexgap I think we should get one of the styling PRs merged into master so that we are all using the same sass directory structure and same files for globals

PR #227, opened 2018-10-24
Add delay before processing message from firebase

andrewgordstewart:
When we've written the app to automatically respond to messages in a certain way, some views only appear for a fraction of a second before transitioning to the next view.

By delaying the processing of an incoming message, the user can at least see this screen momentarily.

### DIFF
--- a/src/redux/message-service/saga.ts
+++ b/src/redux/message-service/saga.ts
@@ -1,5 +1,5 @@
 import { fork, take, call, put, select, actionChannel } from 'redux-saga/effects';
-import { buffers } from 'redux-saga';
+import { buffers, delay } from 'redux-saga';
 import hash from 'object-hash';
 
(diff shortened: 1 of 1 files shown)


src/redux/message-service/saga.ts
```diff
@@ -116,6 +116,9 @@ function* receiveFromFirebaseSaga(address) {
     const message = yield take(channel);
     // tslint:disable-next-line:no-console
     console.log('[MESSAGE SAGA] message received');
+
+    yield delay(500);
```
lalexgap: It'd be nice to be able to easily flag this on and off based on an environment variable for debugging purposes.

lalexgap: Probably a good idea to put a comment here explaining why we have a delay.

tomclose:
Going to close this, as we didn't end up using it for the demo and it doesn't seem like we'd want this going forward.

PR #241, opened 2018-10-26
Basic Withdraw Screen

lalexgap:


src/wallet/components/WalletController.tsx
```diff
@@ -70,11 +74,37 @@ export default class WalletController extends PureComponent<Props> {
           );
         }
         break;
+        case CommonState.WithdrawalComplete:
```
tomclose: The formatting is a bit messed up here.

PR #253, opened 2018-10-28
Validate input when creating new game

andrewgordstewart:
Prevents the following error: 
screen shot 2018-10-28 at 12 54 45 pm

### DIFF
--- a/src/components/CreatingOpenGameModal.tsx
+++ b/src/components/CreatingOpenGameModal.tsx
@@ -11,12 +11,17 @@ interface Props {
   cancelOpenGame: () => void;
 }
 
+const MIN_BUYIN = 0.001;
+const MAX_BUYIN = 1;
+
 export default class CreatingOpenGameModal extends React.PureComponent<Props> {
   buyinInput: any;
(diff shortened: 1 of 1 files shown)


src/components/CreatingOpenGameModal.tsx
```diff
@@ -51,7 +64,11 @@ export default class CreatingOpenGameModal extends React.PureComponent<Props> {
               ref={this.buyinInput}
             />
             <small className="form-text text-muted">
-              Enter an amount between 1 and 0.001
+            {
+              this.invalidBuyin
+              ? `Invalid buy in: ${this.invalidBuyin}. Please enter an amount between ${MIN_BUYIN} and ${MAX_BUYIN}`
+              : `Please enter an amount between ${MIN_BUYIN} and ${MAX_BUYIN}`
```
andrewgordstewart: Should we actually restrict input like so, or should we support smaller input than 0.001 ETH?

tomclose: I think you might also be able to do this at the html level using a number field with min/max. Might be worth a look.

PR #263, opened 2018-10-29
Anonymous login

tomclose:
Still todo - styling isn't great

screen shot 2018-10-29 at 12 29 51 pm

### DIFF
--- a/src/components/ProfilePage.tsx
+++ b/src/components/ProfilePage.tsx
@@ -1,0 +1,84 @@
+import _ from 'lodash';
+import React from 'react';
+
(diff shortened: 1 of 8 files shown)


src/components/ProfilePage.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  handleTwitterChange(e) {
+    this.setState({ twitterHandle: e.target.value });
+  }
+
+
+  updateProfileHandler(e) {
+    e.preventDefault();
+    this.props.updateProfile(this.state.name, this.state.twitterHandle);
```
lalexgap: We should probably do validation somewhere in here

PR #271, opened 2018-10-29
Get Name from the profile instead of login

lalexgap:


src/wallet/containers/WalletContainer.ts
```diff
@@ -10,7 +10,7 @@ const mapStateToProps = (state: SiteState) => {
     challengeState: state.wallet.challenge,
     showWallet: state.wallet.display.showWallet,
     // TODO: We should store this in the wallet state and get it from there
-    loginDisplayName: state.game.gameState ? state.game.gameState.myName : "",
+    loginDisplayName: state.login.profile ? state.login.profile.name : "",
```
andrewgordstewart: Can be `loginDisplayName: state.login.profile || ""`

lalexgap: Can you do that in this case where `state.login.profile` may not be defined and we want to pull the `name` off of it?

PR #273, opened 2018-10-29
Check to make sure code is there before using the contract at address

lalexgap:


src/contracts/SimpleAdjudicator.ts
```diff
@@ -48,7 +49,25 @@ export async function deploySimpleAdjudicator({ channelId, amount }: { channelId
   return await truffleContract.at(deployedContract.options.address);
 }
 
-export async function simpleAdjudicatorAt({ address, amount } : { address: string, amount: BN }) {
+async function verifyContractDeployed(address){
+  // Check if we can access the code at the address if not delay and try again
+  let code = await connectWeb3().eth.getCode(address);
+  let delayAmount = 100;
+  while ((code === '' || code === '0x') && delayAmount < 5000) {
+    await delay(delayAmount);
```
andrewgordstewart: Does this work? `delay` returns an effect that is then run by the middleware. Is that effect just a promise that resolves after `delayAmount`?

lalexgap: Yes, delay returns a promise that can be awaited.

PR #281, opened 2018-10-29
Styling changes

andrewgordstewart:
- The footer bar no longer covers the main screen.
- The message is displayed on the result page.

### DIFF
--- a/src/styles/layout.scss
+++ b/src/styles/layout.scss
@@ -12,7 +12,7 @@ html, body, #root {
   display: flex;
   align-items: center;
   padding-top: 40px;
-  padding-bottom: $navbar-height;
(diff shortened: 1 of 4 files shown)


src/styles/layout.scss
```diff
@@ -11,6 +12,7 @@ html, body, #root {
   display: flex;
   align-items: center;
   padding-top: 40px;
+  padding-bottom: $navbar-height;
```
tomclose: Isn't the `$navbar-height` the height of the very top navbar? (And not the bottom navbar.)

kerzhner: @tomclose Andrew is referring to the footer height here, which is the same as the navbar height

PR #292, opened 2018-11-20
Split external wallet actions into outgoing and incoming

tomclose:
Part of the wallet refactor. Thought it was worth making this switch separate to the main PR, as it's possible to do so.


src/wallet/interface/incoming.ts
```diff
@@ -0,0 +1,139 @@
...
+// FUNDING
+// =======
+
+export const FUNDING_REQUEST = 'WALLET.FUNDING.REQUEST';
+
+export const fundingRequest = (
+  channelId: string,
+  myAddress: string,
+  opponentAddress: string,
+  myBalance: BN,
```
kerzhner: I remember having issues with BN right before DevCon. Do you think that we should start moving away from BN and pass around strings?

src/wallet/interface/incoming.ts
```diff
@@ -0,0 +1,139 @@
...
+  opponentBalance: BN,
+  playerIndex: PlayerIndex,
+) => ({
+  type: FUNDING_REQUEST as typeof FUNDING_REQUEST,
+  channelId,
```
kerzhner: Are types inferred from the `fundingRequest` arguments?

src/wallet/interface/incoming.ts
kerzhner: 👌

src/wallet/interface/incoming.ts
kerzhner: Thanks!

PR #293, opened 2018-11-21
Remove truffle-solidity-loader

lalexgap:
Remove truffle solidity loader (so `yarn start` doesn't need to be called twice)

I've also removed starting up `ganache-cli` in the start script as I was having issues getting it working with the `deploy_contracts` script. 

This means that ganache needs to be started up separately when targeting a development network.

As before the target network can be set by setting `TARGET_NETWORK`  in the `.env.` file. If this isn't set it defaults to `development` for `yarn start` and `ropsten` for `yarn build`.

Possible improvements:
- [x] Start up `ganache-cli` if there's no ganache running and get that working with `deploy_contracts` script.
- [x] Modify `deploy_contracts.js` to output to the console immediately instead of when the whole process completes. (While still blocking until the deploy process is complete)

EDIT: Ganache now starts up if not running, and I've added a new command to start up ganache.


package.json
```diff
@@ -66,17 +66,15 @@
...
+
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "run-s test:truffle 'test:app --all'",
     "test:ci": "run-s build test:truffle 'test:app --all --ci'",
     "test:app": "node scripts/test.js --env=jsdom",
     "test:truffle": "node scripts/test_truffle.js",
-    "truffle:migrate": "node_modules/.bin/truffle migrate --network $npm_package_config_deploymentNetwork",
+    "truffle:migrate": "node -e 'require(\"./scripts/deploy_contracts.js\").deployContracts()'",
```
andrewgordstewart: What's the benefit of having a script that exports a function, then importing/calling that function in a second "script" via `node -e`?

For comparison, the rest of the scripts files are simply scripts that are run.

scripts/start.js
```diff
@@ -86,41 +86,16 @@ choosePort(HOST, DEFAULT_PORT)
       urls.lanUrlForConfig
     );
 
-    //Default accounts to seed so we can have accounts with 1M ether for testing
-    var accounts =
```
andrewgordstewart: Can we at least extract this into a `start_ganache.js` script so that people can continue to use ganache in the same way, via `yarn start:ganache` or something like that?

package.json
lalexgap: I did this for ease of use of calling the logic from other scripts. The main use case of `deploy_contracts` is to be called from the `build` and `start` script.

 We need to call `deploy_contracts` from these scripts to ensure that `process.env.TARGET_NETWORK` has been setup correctly before attempting to deploy.

 It seemed more natural for this to be a function then a script that has to be invoked with a `child_process`.

The `truffle:migrate` command was just updated to maintain compatibility.

Perhaps this would be clearer if the `deploy_contracts.ts` was moved to a helper folder or file within the scripts folder.

andrewgordstewart: Makes sense.

It seems fine in the scripts folder, though it is perhaps a little out of place.

kerzhner:
@lalexgap Should we update the README as well?

PR #295, opened 2018-11-22
Switch from truffle-contract to Ethers.js

lalexgap:
Based on work by @andrewgordstewart I've switched from `truffle-contract` to `ethers.js` for interacting with adjudicator contract.


src/wallet/redux/sagas/blockchain.ts
```diff
@@ -155,38 +156,47 @@ function handleError(action, err) {
   return put(action(message));
 }
 
-function* watchAdjudicator(deployedContract) {
+function convertBigNumberArgsToBN(argumentObject): any {
```
lalexgap: Dealing with different BigNumber libraries is a pain. It'd be nice to find a more maintainable solution.

PR #296, opened 2018-11-22
Format code on all files

lalexgap:
Called format on all our files to clean them up (using https://marketplace.visualstudio.com/items?itemName=jbockle.jbockle-format-files)

We should turn on `format on save` in vs code to keep formatting nice and pretty.
screen shot 2018-11-22 at 11 28 06 am


tomclose:
Closing because it's easy to redo and there are a lot of conflicts with the wallet work.

PR #307, opened 2018-11-26
Add addresses and private keys to the scenarios

tomclose:
I realised I needed valid addresses and private keys for testing inside the wallet.

Eventually we'll probably need wallet-specific scenarios, but this fix will do for now.


kerzhner:
Looks good to me.

PR #308, opened 2018-11-27
Build out the opening stage. Fixes #301.

### LINKED ISSUE #301: Build out the opening stage

tomclose:


src/wallet/redux/reducers/opening.ts
```diff
@@ -1,18 +1,131 @@
...
     default:
       return unreachable(state);
   }
 };
 
-const waitForChannelReducer = (state: WaitForChannel, action: any) => {
-  return state;
+const waitForChannelReducer = (state: states.WaitForChannel, action: actions.WalletAction) => {
+  switch (action.type) {
+    case actions.OWN_POSITION_RECEIVED:
```
kerzhner: @tomclose Not a big deal, but thought I would mention anyways. A lot of the logic in the `OWN_POSITION_RECEIVED` and `OPPONENT_POSITION_RECEIVED` cases is very similar. Is there a way to factor it out into a more generic function?

src/wallet/redux/reducers/__tests__/opening.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  bsPrivateKey,
+  channelId,
+  participants,
+  channelNonce,
+} = scenarios.standard;
+
+const defaults = { uid: 'uid', displayMode: states.DisplayMode.None, address: asAddress, privateKey: asPrivateKey };
+
+describe('when in WaitForChannel', () => {
+  describe('when we send in a PreFundSetupA', () => { // preFundSetupA is A's move, so in this case we need to be player A
```
kerzhner: @tomclose These tests look great! Have we thought about how we are going to test for side effects of state transitions?

src/wallet/redux/reducers/opening.ts
tomclose: I separate out some of the functionality in https://github.com/magmo/rps-poc/pull/310 but agree this could still be better. Plan was to wait until there are more reducers, so it's clearer what the common parts are.

src/wallet/redux/reducers/__tests__/opening.test.ts
tomclose: Currently the only side-effects are (a) sending a message to your opponent and (b) sending a transaction to the blockchain. In reducer-land those side effects are both recorded in the state:

https://github.com/magmo/rps-poc/blob/wallet/src/wallet/states/shared.ts#L9-L10

This means that to check side-effects you just check the state.

PR #310, opened 2018-11-27
Running stage

tomclose:
Includes https://github.com/magmo/rps-poc/pull/308.


src/wallet/states/responding.ts
```diff
@@ -15,6 +15,9 @@ interface AcknowledgeChallenge extends AdjudicatorExists {
   type: typeof ACKNOWLEDGE_CHALLENGE;
   stage: typeof RESPONDING;
 }
+export function acknowledgeChallenge<T extends AdjudicatorExists>(params: T): AcknowledgeChallenge {
+  return { type: ACKNOWLEDGE_CHALLENGE, stage: RESPONDING, ...adjudicatorExists(params) };
```
lalexgap: Should we be setting displayMode here so the wallet will pop up for the user to acknowledge the challenge?

tomclose: Yes. Good spot.

PR #314, opened 2018-11-28
Wallet Challenging and Responding Refactor/Scaffolding

lalexgap:
Scaffold out the states and actions for challenging and responding. There's still some work to do to properly handle submitted the challenge that will be tackled as part of #311 .  Addresses #300 #302.


src/wallet/redux/reducers/__tests__/challenging.test.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+
+describe('when in APPROVE_CHALLENGE', () => {
+    const state = states.approveChallenge({ ...defaults });
+    describe('when a challenge is approved', () => {
+        const action = actions.approveChallenge();
+        const updatedState = walletReducer(state, action);
+        itTransitionsToStateType(states.INITIATE_CHALLENGE, updatedState);
+        itSendsATransaction(updatedState);
```
tomclose: @lalexgap Your browser seems to still default to 4 spaces instead of two .. :(

src/wallet/redux/reducers/responding.ts
```diff
@@ -1,6 +1,100 @@
 import { WalletState, RespondingState } from '../../states';
+import * as states from '../../states/responding';
+import * as challengeStates from '../../states/challenging';
+import * as runningStates from '../../states/running';
```
tomclose: Are you aware that these states are all exported from `../../states`, so you could just do the following if you wanted?:
```ts
import * as states from '../../states'
```

src/wallet/states/responding.ts
```diff
@@ -6,49 +6,73 @@ export const RESPONDING = 'RESPONDING';
 // state types
 export const ACKNOWLEDGE_CHALLENGE = 'ACKNOWLEDGE_CHALLENGE';
 export const CHOOSE_RESPONSE = 'CHOOSE_RESPONSE';
-export const ACKNOWLEDGE_MUST_TAKE_MOVE = 'ACKNOWELDGE_MUST_TAKE_MOVE';
+export const SELECT_CHALLENGE_RESPONSE = 'SELECT_CHALLENGE_RESPONSE';
```
tomclose: I see you now have `SELECT_CHALLENGE_RESPONSE` and `CHOOSE_RESPONSE`. It isn't clear to me what the difference between the two of these are. Do we definitely need both of them?

lalexgap: Hmm, the naming should probably be clarified. `SELECT_CHALLENGE_RESPONSE` should probably be `SELECT_RESPOND_WITH_MOVE`

tomclose: Presumably you the move you pick would be part of the `CHOOSE_RESPONSE` e.g.
```ts
enum Response { RespondWithMove, Refute, RespondWithAlternativeMove }

chooseResponse(response: Response) { ... }
```

lalexgap: @tomclose Right, sorry I think I was thinking of actions. I just realized that `SELECT_CHALLENGE_RESPONSE` is actually not used, so I must of introduced it and forgot about it. :)

src/wallet/redux/reducers/responding.ts
lalexgap: I think I did this for clarity in development so that `states` would only contain the Responding states, and it would be clear when I was returning a state that is not a `responding` state

src/wallet/redux/reducers/__tests__/challenging.test.ts
lalexgap: Hmm, do you only get 2 when you run auto-format for describe blocks?

tomclose: I have `"editor.tabSize": 2,` in my user settings.

tomclose: How do I run auto-format?

lalexgap: Looks like my tabSize was set to 4. Had to set to 2, turn off auto-detect, and then run auto-format.

PR #329, opened 2018-11-30
Get something in place for most of the screens

tomclose:
Only the responding screens are left to do. The other screens are only a v0 though.


kerzhner:
Looks good to me 👍

PR #337, opened 2018-12-03
Run transaction tests on CI

lalexgap:
Some small tweaks so that the new transactions tests run from the command `test:app` and `test:ci`


kerzhner:
Nice! LGTM. Do you know where the whitespace changes are coming from?

lalexgap:
> Nice! LGTM. Do you know where the whitespace changes are coming from?

I've turned on autoformat on save, and I don't think we've ever run format on some of our files 😄

PR #340, opened 2018-12-05
Refactor wallet screens

tomclose:
I've extracted some common screens into shared components. Still work to be done here but it should make it easier to change the look consistently when we need to.


src/wallet/components/AcknowledgeX.tsx
```diff
@@ -1,26 +1,30 @@
...
 interface Props {
-  challengeResponseAcknowledged: () => void;
+  title: string;
+  action: () => void;
+  actionTitle: string;
+  description: string;
 }
 
-export default class AcknowledgeChallengeResponse extends React.PureComponent<Props> {
+export default class AcknowledgeX extends React.PureComponent<Props> {
```
lalexgap: I wonder if we should call these `Acknowledge` and move them into a subfolder to indicate they're generic components instead of `AcknowledgeX`?

PR #347, opened 2018-12-06
Wallet to app messaging

kerzhner:


src/wallet/redux/sagas/message-sender.ts
```diff
@@ -1,4 +1,6 @@
+import { put } from "redux-saga/effects";
+import { ResponseAction } from "../../interface/outgoing";
 
-export function* messageSender(message) {
-  // todo
+export function* messageSender(action: ResponseAction) {
```
lalexgap: We'll probably need to clear out the messageOutbox after sending a message or we'll be sending a  message everytime an action is dispatched (and possibly get stuck in a infinite loop with `saga-manager`?)

kerzhner: Yeah, good call on that.

kerzhner: https://github.com/magmo/rps-poc/pull/347/commits/d94ef53b14b44bce8d0701d8c194412da0d88a8d

PR #355, opened 2018-12-13
Wallet Funding: non happy path + Removing post fund states from game reducer

kerzhner:
Non happy path covers:
1. Player rejects funding in the Approve Funding screen.
1. Player rejects deploy transaction.
1. Player rejects deposit transaction.

Now, both players are able to get into game play.

To be done in a later PR:
A player is not notified when the opponent rejects funding for any of the reasons above.


src/wallet/redux/reducers/funding.ts
```diff
@@ -184,8 +194,23 @@ const bWaitForDepositToBeSentToMetaMaskReducer = (state: states.BWaitForDepositT
 
 const bSubmitDepositInMetaMaskReducer = (state: states.BSubmitDepositInMetaMask, action: actions.WalletAction) => {
   switch (action.type) {
+    // This case should not happen in theory, but it does in practice.
```
lalexgap: Its crazy that this can happen before transaction submitted

kerzhner: @lalexgap This might not be a realistic scenario when using Ropsten or the mainchain, but it does happen locally with ganache.

Issue #348, opened 2018-12-06
Split out integration tests


tomclose:
They need to run sequentially

PR #367, opened 2019-01-08
Add separate config files for integration vs app tests

lalexgap:
I've added two separate config files for unit vs integration tests. 

Addresses #348 

This shouldn't be merged until after the wallet branch is merged into master.


lalexgap:
Will recreate for the monorepo

